PR #26, opened 2020-11-24
Return an object with page property

toomuchdesign:

Feature

`getPage` returns the page element, preventing method from returning additional info.

## What is the new behaviour?

`getPage` returns an object with a `page` field holding the page element. The returning object will possibly be extended with additional info.

Yes:

```js
const myPage = getPage({ route: '/my-route' });
```
...becomes
```js
const { page: myPage } = getPage({ route: '/my-route' });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,11 @@ import { getPage } from 'next-page-tester';
 
 describe('Blog page', () => {
   it('renders blog page', async () => {
-    const Page = await getPage({
+    const { page } = await getPage({
       route: '/blog/1',
     });
 
-    render(Page);
+    render(page);
     expect(screen.getByText('Blog')).toBeInTheDocument();
 
     fireEvent.click(screen.getByText('Link'));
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -14,11 +14,11 @@ import { getPage } from 'next-page-tester';
 
 describe('Blog page', () => {
   it('renders blog page', async () => {
-    const Page = await getPage({
+    const { page } = await getPage({
```
toomuchdesign, line 17: Hi @Meemaw,
quick help request :)

How do you would be a good name for the property holding the returned page element? Consider that it might be a page element or the `App` element wrapping the page.

Candidates:
- page
- element
- pageElement
- anything else?

Meemaw: Hmmm, I think I would stay with `page` for now. Anyway those are always hard to get right the first time so I wouldn't worry too much about it. I think with usage and adoption cleaner API will emerge if needed.

Btw, unrelated question regarding the `App` element. If I understand correctly it is not included by default. How come?

toomuchdesign: Let's stick to `page`, then. Thanks 🙌 

I set the app option to false by default just because I wanted to be defensive. I'll default it to true with v.0.8.0.

Issue #29, opened 2020-11-25
Wrap page with document


Meemaw:
## 🚀 Feature Proposal

Wrap page with document (as it is currently done with App).

## Motivation

NextJS supports [custom `Document`](https://nextjs.org/docs/advanced-features/custom-document). There might be some involved logic there to render things (e.g. header meta attributes for SEO).

## Example

```ts
import { render, screen, fireEvent } from '@testing-library/react';
import { getPage } from 'next-page-tester';

describe('Blog page', () => {
  it('renders blog page', async () => {
    const { page } = await getPage({
      route: '/blog/1',
    });

    const { container } = render(page);
    expect(
      container.querySelector('meta[description="Description"]')
    ).toHaveAttribute('content', 'My custom description');
  });
});
```

Meemaw:
I just realised this wain mentioned in todos already: https://github.com/toomuchdesign/next-page-tester#todos

Do you mind if I work on this @toomuchdesign ?

toomuchdesign:
It'd be amazing, you're definitely welcome!
I have never ensured that basic Next Document component is actually renderable in JSDOM environment. Let's see. :)

Feel free to ping me for any question about the current codebase. ✌️

PR #30, opened 2020-11-25
feat: support custom document

Meemaw:

Feature, fixes https://github.com/toomuchdesign/next-page-tester/issues/29.

`_document` is ignored.

## What is the new behaviour?

`_document` is included in the rendered markup.

No


src/__tests__/__fixtures__/pages/_document.js
```diff
@@ -1,4 +0,0 @@
-import React from 'react';
```
Meemaw, line 1: Incorrect `_document` file -- was breaking the tests.

toomuchdesign:
Thanks @Meemaw, looks like it was harder then expected.
Would you briefly explain me how the Document component (custom or default) is handled by Next? I saw there's a context provided involved...

Meemaw:
So document Component by default renders the following (this it the minimum required, and you should not remove stuff from there, just add it)

```js
  render() {
    return (
      <Html>
        <Head />
        <body>
          <Main />
          <NextScript />
        </body>
      </Html>
    )
  }
```

1. `Html` components just renders `html` element with some extra attributes based on locale,... etc
2. `Head` component is quite complicated and will in Next.JS render a lot of stuff (css, assets, preloading links, etc,...). I think those are not relevant to us so and to the user testing, so we are pretty much just using it as a children passthrough.
3. `NextScript` will render JS chunks associated with a page -- not relevant to us
4. `Main` is where the magic happens: https://github.com/vercel/next.js/blob/2790ea411b8b0bf2be627a5cc5ba77626079c46d/packages/next/pages/_document.tsx#L524 -- see that html is read from the `DocumentContext` therefore we need to wrap it

I've actually changed the implementation a bit to re-use some of the NextJS methods to abstract the Provider away from us. Should be also more future-proof. I think it is cleaner now and should make more sense @toomuchdesign

Meemaw:
Open question -- should we alway render `Document` component, even if user does not have a custom one defined? Implementation wise that is trivial as we can just use the default implementation: https://github.com/vercel/next.js/blob/canary/packages/next/pages/_document.tsx#L64

The rendered markup would more closely resemble that is the actual output of the NextJS rendered page. In that cause I would rename the `useCustomDocument` option to `useDocument`.

toomuchdesign:
Hi @Meemaw, that's exactly what I thought when I saw your PR. I'm still gathering question/comments, but I'd say `useDocument` is the way to go.

Do you think it would also apply to `useCustomApp` or Next.js has no idea of default app component at all?

src/__tests__/custom-document/__fixtures__/custom-document-with-gip/pages/_app.js
```diff
@@ -0,0 +1,20 @@
...
+  );
+}
+
+CustomApp.getInitialProps = async (appContext) => {
+  return {
+    pageProps: {
+      ctx: appContext,
+      fromCustomApp: true,
+      propNameCollision: 'from-app',
+    },
```
toomuchdesign, line 18: Probably we don't want to test that custom App is called with expected props here. Do you think we can safely remove these props?

toomuchdesign (changes requested):
Hi @Meemaw, great job here. I didn't expect custom Document logic to be so convoluted.
I left a few comments, make sure they make sense in your context.

As for the high level picture I agree with providing a generic `useDocument` option, and provide custom document in case the project doesn't have one.

Do you think we can safely ship this new feature enabled by default? If we're not 99% sure we could consider keeping the flag disabled at first. I expect this utility to be especially used in relatively simple projects.

Few last considerations:
- as a first iteration, we might also consider releasing a subset of the features currently supported by custom Document
- does the current implementation support [`next/head` in-page](https://nextjs.org/docs/api-reference/next/head)?
- Is there anything mentioned in [Document/caveats](https://nextjs.org/docs/advanced-features/custom-document#caveats) we should keep in consideration?

Meemaw:
Thanks for the review @toomuchdesign. I did some cleanups. Current state:
- Prop renamed to `useDocument` which will fallback to `Document's` default implementation
- `useDocument` disabled by default
- Using `<Head>` element in the page will fail (its a bit tricky, I'm having some issues getting it to work)

toomuchdesign:
Let's move this branch into the original repo so that we can both push.
Push forces not allowed anymore :)

toomuchdesign:
Hi @Meemaw,
I reviewed the tests (and added one). I hoped I was able to get your testing intention, there. I stumbled upon a weird issue [here](https://github.com/toomuchdesign/next-page-tester/pull/30/files#diff-e190b4f91881af0e53120e9539ae4d3084ec4e110a192b3efe6ef46124ef6aa2R29), though.

The comparison between the expected element and the element extracted from Next.js output makes Jest throw an error somewhere about circular reference. I'm still trying to figure out why. I temporarily disabled the concerned assertion.

src/_document/getDocumentInitialProps.tsx
```diff
@@ -0,0 +1,24 @@
...
+  Document,
+  pageObject,
+  renderPage,
+}: {
+  Document: DocumentType;
+  pageObject: PageObject;
+  renderPage: RenderPage;
+}): Promise<DocumentInitialProps> {
+  const getDocumentInitialProps =
+    Document.getInitialProps || NextDocument.getInitialProps;
```
toomuchdesign, line 16: Do we need this fallback to `NextDocument.getInitialProps` ? If I'm not mistaken custom Document object should extend Next.js `Document` which is the same class we use here as fallback.

Meemaw: I think `getInitialProps` is typed as optional in type definitions, so we do a fallback to avoid having to do checks.

src/__tests__/custom-document/__fixtures__/special-extension/pages/_document.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import Document, { Html, Head, Main, NextScript } from 'next/document';
+
+class CustomDocument extends Document {
+  render() {
+    return (
+      <>
```
Meemaw, line 7: I'm not sure if omitting `<Html>` is valid. Have you tried using such document in Next?

src/__tests__/custom-document/custom-document.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+        const { container: expected } = render(
+          <CustomApp Component={CustomDocumentWithGIP_Page} pageProps={{}} />,
+          { container: document.createElement('html') }
+        );
+
+        // @NOTE this match produces a circular structure error
+        // https://github.com/facebook/jest/issues/10577
+        // expect(actual).toEqual(expected);
+        expect(true).toBe(true);
```
Meemaw, line 36: Does this happen if you remove the custom 'html' container as well? 🤔

src/_document/getDocumentInitialProps.tsx
toomuchdesign: 👍 It seems that the fallback never happen on runtime, but TS complains about it. Skipped that line for code coverage.

src/__tests__/custom-document/custom-document.test.tsx
toomuchdesign: No! It seems to only happen with the element extracted from `#__next` element. With `container` it works fine.

I also noticed that the element has an unexpected comment created by Next.js:
```js
<div>
  _app
  <!-- -->
  /custom-document-with-gip/page
</div>
```

src/__tests__/custom-document/custom-document.test.tsx
```diff
@@ -0,0 +1,109 @@
+import React from 'react';
+import '@testing-library/jest-dom';
+import { render, screen } from '@testing-library/react';
+import { getPage } from '../../index';
+import CustomDocumentWithGIP_Page from './__fixtures__/custom-document-with-gip/pages/page';
+import CustomApp from './__fixtures__/custom-document-with-gip/pages/_app';
+
+beforeEach(() => {
+  // Suppress validateDOMNesting error logs
```
Meemaw, line 9: Should probably document this; -- both the error and the workaround.

toomuchdesign: You're right. I didn't realise final users would get the same issue. :)

toomuchdesign:
I we find out what goes on in `custom-document.test.tsx`, we might consider merging the PR.
I marked `useDocument` option as experimental.

How do you see it @Meemaw?

Meemaw:
Agree, it's definitely experimental, but it is something we can build upon.

I'll look into the problematic test later today.

src/__tests__/use-document/__fixtures__/custom-document-with-gip/pages/page.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
 
 export default function CustomDocumentWithGIP_Page() {
-  return <>/custom-document-with-gip/page</>;
+  return <span>/custom-document-with-gip/page</span>;
```
Meemaw, line 4: Using `<span>` otherwise `expect(actual).toEqual(expected);` would fail due to `renderToStatic*` behaviour which will somehow truncate strings -- this would not affect actual page looks so it's not really an issue.

src/_document/render.ts
```diff
@@ -29,7 +29,7 @@ export default async function renderDocument({
 
   const renderPage: RenderPage = () => {
     // Render markup that will be injected into #__next element by Next.JS in the Main component of Document
-    const html = renderToString(pageElement);
+    const html = renderToStaticMarkup(pageElement);
```
Meemaw, line 32: This removes the extra comment that appeared in the output.

Meemaw:
@toomuchdesign should be good to go

toomuchdesign (approved):

toomuchdesign:
Let's merge! Thanks for this funny trip into Next.js crazy internals, @Meemaw. 😎

PR #41, opened 2020-11-29
fix: correctly handle <Head /> in pages & client side events in document

Meemaw:

Bug fix

- Rendering `<Head><title>Something</title></Head>` inside `pages` will do nothing.
- Client side events do not work when `useDocument=true`

## What is the new behaviour?

- Rendering  `<Head><title>Something</title></Head>` inside `pages` will correctly render <title> in document
- Client side events work when `useDocument=true`

No


src/__tests__/real-world-example/real-world-example.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+    expect(head.querySelector('title')?.textContent).toEqual('Create Next App');
+
+    // Correctly passes query
+    expect(screen.getByText('Hello Matthew')).toBeInTheDocument();
+
+    userEvent.click(screen.getByText('To page A'));
+
+    // TODO: fix client side navigation when useDocument: true
+    // await screen.findByText('Back to root');
```
Meemaw, line 34: Client side navigation does not work when `useDocument: true` -- TODO: investigate

package.json
```diff
@@ -84,7 +86,10 @@
     "coverageReporters": [
       "text",
       "lcov"
-    ]
+    ],
+    "moduleNameMapper": {
+      "\\.(css|less|scss)$": "identity-obj-proxy"
```
Meemaw, line 91: So `jest` wont crash when pages are importing CSS files

src/_document/render.tsx
```diff
@@ -22,14 +24,22 @@ export default async function renderDocument({
...
-  // See: https://github.com/vercel/next.js/blob/2790ea411b8b0bf2be627a5cc5ba77626079c46d/packages/next/next-server/server/render.tsx#L563
-  // Problem, can we get this to evaluate to false: https://github.com/vercel/next.js/blob/2790ea411b8b0bf2be627a5cc5ba77626079c46d/packages/next/next-server/lib/side-effect.tsx#L3?
-  const head: JSX.Element[] = [];
+  let head: JSX.Element[] = [];
 
   const renderPage: RenderPage = () => {
     // Render markup that will be injected into #__next element by Next.JS in the Main component of Document
-    const html = renderToStaticMarkup(pageElement);
+    const html = renderToStaticMarkup(
+      <HeadManagerContext.Provider
```
Meemaw, line 32: Implementation from https://github.com/vercel/next.js/blob/canary/packages/next/next-server/server/render.tsx#L561

src/loadPage.ts
```diff
@@ -10,8 +10,22 @@ export function loadPage({
...
   pagePath: string;
 }): NextPageFile {
+  const tmpWindow = global.window;
+
+  // Next.JS executes some rendering logic conditionally, depending if its on server.
+  // https://github.com/vercel/next.js/blob/canary/packages/next/next-server/lib/side-effect.tsx#L3
+  // We want that logic to execute, so we have to trick it to think code is executing on server
+  // Because expression is evaluated on top lelve, we have to "mock" it before module is evaluated -- before require;
+  // @ts-ignore
+  delete global.window;
```
Meemaw, line 20: Hacky, but I don't see any better way BTH

src/__tests__/real-world-example/real-world-example.test.tsx
toomuchdesign: I haven't had time to verify, but I guess this could be at least partially related to the way we wrap the Pafe (including `Document` which is written to be rendered server-side, only) with `RouterProvider` and `NavigationProvider`.

Client navigation is currently implemented as follows:
- `RouterProvider` wraps the page and makes available a the router mock for current page via context
- router mock is configured to issue a router mock update on `route.push`/`route.replace`
- Below `RouterProvider` lives `NavigationProvider` which invokes `makePage` each time router object changes

`makePage` creates a new `pageObject` based on the new router and calls `makePageElement` which recreates the whole Document + App + Page element from scratch.

All this happens on React runtime.

Meemaw: Okay, I have come up with a way to make it work -- more hacks :)

The problem is that as soon as you render elements on server using `renderToStatic*` function scope is lost it won't work as expected in the tests.

Will prepare changes later today.

Meemaw: Should be good to go @toomuchdesign 🚀

src/__tests__/real-world-example/__fixtures__/README.md
```diff
@@ -0,0 +1,30 @@
+This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
```
toomuchdesign, line 1: Is the default readme left here intentionally?

src/__tests__/real-world-example/__fixtures__/next-env.d.ts
```diff
@@ -0,0 +1,2 @@
+/// <reference types="next" />
```
toomuchdesign, line 1: Is it needed for testing purpose?

src/_document/render.tsx
```diff
@@ -22,17 +27,36 @@ export default async function renderDocument({
...
 
+  // https://github.com/vercel/next.js/blob/canary/packages/next/pages/_document.tsx#L524
+  // Default behaviour of Main component is to dangerouslySetInnerHTML with the html
+  // string rendered above. This works, but will break all client side interactions
+  // as event handlers are lost in static markup
+  // @ts-ignore
+  documentModule.Main = () => {
+    const { inAmpMode, docComponentsRendered } = useContext(DocumentContext);
+    docComponentsRendered.Main = true;
+    if (inAmpMode) return <>{AMP_RENDER_TARGET}</>;
```
toomuchdesign, line 56: Are these 3 lines (54,55,56) necessary for the scope of our tests?
I can remove them and still have green tests.

src/_document/render.tsx
```diff
@@ -22,17 +27,36 @@ export default async function renderDocument({
...
+          mountedInstances: new Set(),
+        }}
+      >
+        {pageElement}
+      </HeadManagerContext.Provider>
+    );
     return { html, head };
   };
 
+  // https://github.com/vercel/next.js/blob/canary/packages/next/pages/_document.tsx#L524
```
toomuchdesign, line 48: I'd replace this link (and any other link pointing to git repos' files) with:
https://github.com/vercel/next.js/blob/v10.0.3/packages/next/pages/_document.tsx#L524

..in order to make sure we don't lose reference to the relevant code when target repo changes. :)

src/loadPage.ts
toomuchdesign: I actually like it. It makes perfectly sense since we're testing isomorphic code in a polyfilled environment. The only better solution I can think of would consist in somehow temporarily disabling JSDOM or delaying its initialisation but it seems not feasible.

toomuchdesign (commented):
I left a few comments but it seems all quite fine. Rendering the document component is JSDOM seems quite a painful accomplishment. :)

Real world example tests are generating a few more `validateDOMNesting` errors. We might suppress them repeating what we did here:
https://github.com/toomuchdesign/next-page-tester/blob/master/src/__tests__/use-document/use-document.test.tsx#L8

..or consider providing this mock to all tests as global jest configuration.

PS. What was eventually the issue with document rendering + client side navigation?

src/__tests__/real-world-example/__fixtures__/README.md
Meemaw: Yes, I have generated app with `npx create-next-app`; idea is that I'm able to go into that directory and run `yarn next dev` and compare `html` from server with what we render.

src/__tests__/real-world-example/__fixtures__/next-env.d.ts
Meemaw: Related to previous answer.

src/_document/render.tsx
Meemaw: They are not needed, but wanted to have an implementation that matches NextJS

src/_document/render.tsx
Meemaw: Good idea -- done;

Meemaw:
@toomuchdesign addressed the comments. Also added global jest setup file that should get rid of these warning for all files.

> PS. What was eventually the issue with document rendering + client side navigation?

I tried to explain it here: https://github.com/toomuchdesign/next-page-tester/pull/41/files#diff-3483f576877de2f3a21d64653dea24f689cbdd0ce46dbf00f545dcd4b743667bR49

It seems client event handlers are lost when using `dangerouslySetInnerHTML` which `NextJS` uses server-side.

Related question: https://stackoverflow.com/questions/54109790/how-to-add-onclick-event-to-a-string-rendered-by-dangerouslysetinnerhtml-in-reac

src/__tests__/real-world-example/real-world-example.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+    userEvent.click(screen.getByText('Click me'));
+    expect(screen.getByText('Count: 1')).toBeInTheDocument();
+
+    // Make sure cleint navigation work
+    userEvent.click(screen.getByText('To page A'));
+
+    const backToRootLink = await screen.findByText('Back to root');
+
+    // TODO: something like this should work
+    // screen.getByText('Came from http://localhost:3000/');
```
Meemaw, line 42: This is something that does not work yet. On client side navigation we should set `referer` header on req.

src/_document/render.tsx
toomuchdesign: I see. In this case I'd friendly suggest to:
- make clear in a comment that those lines do nothing
- add an `/* istanbul ignore next */` before line 56 to restore code coverage to 100%

I'm insisting in keeping test coverage to 100% to pre-catch some of the many regression bug this project will suffer.

toomuchdesign (changes requested):
Left just one more change request, if you agree.

I'll push a commit to mock `window.scrollTo` and prevent "not implemented" errors

src/__tests__/real-world-example/real-world-example.test.tsx
Meemaw: TBH, I was just testing how it works in browser (with NextJS server) and wanted to match the behaviour in tests. Perhaps this is out of scope of this project, but we could implement it as we are in control of the HTTP req. User could not mock this one out, because this is after initial render.

toomuchdesign (approved):

toomuchdesign:
I tried to figure out how `Head` components output get updated during client navigation.

I understood that the `head` element gets updated because `NavigationProvider` re-renders the whole page element, including `document`, is that right?

Just talking loud, in future we might consider trying to setup a `HeadManagerContext` provider [as happens in Next.js](https://github.com/vercel/next.js/blob/c8cd77a856248346768042f4553a8513195311a4/packages/next/next-server/server/render.tsx#L568) and like we currently do with router.

Tomorrow I'm going to merge this, feel free to ping me if you still wanted to work on this PR. ✌️

Meemaw:
Yes we rerende whole page on client navigation.

Good to go from my side!

PR #44, opened 2020-12-03
fix: client side window

Meemaw:

Bug fix

Using libraries that rely on `window` object, e.g. `react-hook-form` (https://github.com/react-hook-form/react-hook-form/blob/398bb117b6dac0e9d4c5bc2d7c2ed328c932332e/src/useForm.ts#L77) won't work

## What is the new behaviour?

Using libraries that rely on `window` object works.

No

### DIFF
--- a/src/loadPage.ts
+++ b/src/loadPage.ts
@@ -3,7 +3,7 @@ import fastGlob from 'fast-glob';
 import normalizePath from 'normalize-path';
 import { NextPageFile, ExtendedOptions } from './commonTypes';
 
-function trickNextJsIsServer() {
+function resolveNextJsModulesAsIfOnServer() {
   const tmpWindow = global.window;
 
   // Next.JS executes some rendering logic conditionally, depending if its on server.
@@ -22,11 +22,18 @@ function trickNextJsIsServer() {
 export function loadPage({
   pagesDirectory,
   pagePath,
+  useDocument,
 }: {
   pagesDirectory: string;
   pagePath: string;
+  useDocument: boolean;
 }): NextPageFile {
-  trickNextJsIsServer();
+  // Even though there are places in code where this code would make more sense, it has to be called
+  // before the page is loaded. The problem is that once the page is loaded, all modules get resolved (including NextJS)
+  // and we cannot influence their top level expressions anymore
+  if (useDocument) {
+    resolveNextJsModulesAsIfOnServer();
+  }
 
   // @NOTE Here we have to remove pagePath's trailing "/"
   return require(path.resolve(pagesDirectory, pagePath.substring(1)));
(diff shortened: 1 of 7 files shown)


src/loadPage.ts
```diff
@@ -19,13 +13,23 @@ export function loadPage({
   // @ts-ignore
   delete global.window;
 
-  // @NOTE Here we have to remove pagePath's trailing "/"
-  const page = require(path.resolve(pagesDirectory, pagePath.substring(1)));
+  require('next/dist/next-server/lib/side-effect');
```
Meemaw, line 16: Before requiring the whole page, we require only the module that we want to trick to think is on server.

toomuchdesign:
Hi @Meemaw,
how often do we need `loadPage` to fake a non-browser environment?

Could adding a `nonBrowserEnvironment` (sorry for the bad naming) work, to temporarily make `global.window` unavailable when we load certain files (like custom `_document`)?

Meemaw:
@toomuchdesign  I'm not sure if I understand what you mean.

We need to fake non-browser environment for NextJS to properly render page when `useDocument=true` (head tags).

Previously we did that before loading the whole page, but that was problematic, because it meant window was undefined also for all the modules and external libraries that page is using.

New behaviour is to isolate that only to apply to the specific NextJS module `next/dist/next-server/lib/side-effect`.

toomuchdesign:
I see, thanks @Meemaw!
I checked out the PR and it seems that `trickNextJsIsServer` is not particularly related to `getPage`. It's just Next which needs `next/dist/next-server/lib/side-effect` to be evaluated in non-browser mode before the document is rendered.

What if we moved `trickNextJsIsServer` somewhere into `renderDocument` and maybe rename it to something like `evaluateSideEffect` or whatever that file does?

As bonus we would perform this nasty trick only when `useDocument` set to true and leave the other use-case untouched.

Meemaw:
Agree, it is not related to `getPage`. However, we cannot move it to some step after this because after the `loadPage` is required in the next step it is already too late. At that point all modules are imported and we cannot influence their top level expressions anymore.

We can however add a check to call this only if `useDocument=true` and add a comment why we call it at this point.

toomuchdesign:
If it's needed to be loaded before page is fetched we might:
- assume it's `loadPage` responsibility (and therefore add a flag which we set to `true` when `useDocument` is `true`)
- assume it should not go in `loadPage` and trigger the require trick [somewhere here](https://github.com/toomuchdesign/next-page-tester/blob/fix-client-side-window/src/getPageObject.ts#L14) in `getPageObject` object

How do you see it?

Meemaw:
I did come changes in https://github.com/toomuchdesign/next-page-tester/pull/44/commits/98939a777a5b9517d47ecbfe77892b5691e0083f so we are conditionally loading it. I still think it is actually a good place to execute it as it is related to module loading/resolving.

toomuchdesign (approved):
Bien sûr, looks good to me :)

PR #52, opened 2020-12-07
fix: default document user events

Meemaw:

Bug fix

Client side events are not working when `useDocument=true` and not custom `_document` is not present (cases where we did fallback to NextJS default `Document` implementation)

## What is the new behaviour?

Client side events work in those cases

No


Meemaw:
Im suspecting this might also fix flaky tests but Im not sure.

toomuchdesign (approved):
Ok, I can more or less get what's going on here. Thanks @Meemaw.

Document DOM rendering is requiring quite a lot of maintenance efforts, let's see what's going to happen with the next major releases.

Meemaw:
Yeah, `_document` brings quite some complexity. I think with this PR we are in a much better place. Hopefully they don't plan to break it too much in next major releases. 🙁

PR #54, opened 2020-12-09
Reorganise tests

toomuchdesign:

Reorganise tests in feature folders

No


toomuchdesign:
Hey @Meemaw,
I checked the the filing test output end it might be related to the way we interact with `testing-library` and JSDOM when we render `html/head/body` elements.

This is the JSDOM DOM structure when test fails:

```html
<body>
    <div>
        <html
            lang="en"
        >
            <head>
            [...]
            </head>
            <body>
            [index page, count 1]
            </body>
        </html>
    </div>
</body>

<html>
    <head />
    <body>
        <div>
            <html
            lang="en"
            >
                <head>
                [...]
                </head>
                <body>
                [index page, count 1]
                </body>
            </html>
        </div>
    </body>
</html>

<body>
    <div>
    <html
        lang="en"
    >
        <head>
        [...]
        </head>
        <body>
        [index page, count 1]
        </body>
    </html>
    </div>
</body>
```

Meemaw:
Interesting. Given the flaky nature, it almost feels like DOM is not properly cleaned up and it leaks into next test?

Did you manage to reproduce it locally by any chance? Its a bit hard to debug without that.

toomuchdesign:
Still not, but I've stumbled upon issues between `useDocument` and client navigation. Still investigating... 👀  `client-navigation` tests seem to break by enabling `useDocument`.

toomuchdesign:
When enabling `useDocument` in `client-navigation` tests, it seems that the `router` object becomes `null` after the first client navigation re-render.

I suspect it might deal with the way we currently wrap the page with `RouterProvider` and `NavigationProvider`. Currently the full page is rendered like this:

```js
<RouterProvider>
  <NavigationProvider>
    <Document>
      <App>
        <Page>
      </App>
    </Document>
  </NavigationProvider>
</RouterProvider>
```
...which means that we re-render the custom Document runtime in `NavigationProvider`. I suspect custom <Document> is designed to be rendered server-side once and then selectively updated with its own context. (Please close both eyes for the inaccuracies I just wrote 😄 ).

The following composition makes `client-navigation` green but breaks `real-world-example` tests since `head` element doesn't update.

```js
<Document>
  <RouterProvider>
    <NavigationProvider>
      <App>
        <Page>
      </App>
    </NavigationProvider>
  </RouterProvider>
</Document>
```
I should probably make this comment an issue and move the discussion there.

Meemaw:
Yeah, let's create an issue and I can look into this. It would be great to have a reproducer (maybe just the test that fails when setting `useDocument=true`).

Issue #55, opened 2020-12-10
Client side navigation not working when `useDocument` enabled


toomuchdesign:
## 🐛 Bug Report

`client-navigation` tests break if `useDocument` option is enabled. Tests throw 2 errors:

-  Warning: You seem to have overlapping act() calls, this is not supported. Be sure to await previous act() calls before making a new one.
- `router` being null after navigation rerender

## To Reproduce

Add `useDocument: true` option to the first `getPage` call in [`client-navigation` tests](https://github.com/toomuchdesign/next-page-tester/blob/8639f9fb04635ea7662bae9162997c63c7c6d643/src/__tests__/client-navigation/client-navigation.test.tsx#L16).

## Expected behavior

Even though `useDocument` is an experimental option, we expect it to work with client side navigation.

## Extra note

This might be related with the issue of CI's randomly failing tests

toomuchdesign:
When enabling `useDocument` in `client-navigation` tests, it seems that the `router` object becomes `null` after the first client navigation re-render.

I suspect it might deal with the way we currently wrap the page with `RouterProvider` and `NavigationProvider`. Currently the full page is rendered like this:

```js
<RouterProvider>
  <NavigationProvider>
    <Document>
      <App>
        <Page>
      </App>
    </Document>
  </NavigationProvider>
</RouterProvider>
```
...which means that we re-render the custom Document runtime in `NavigationProvider`. I suspect `<Document>` is designed to be rendered server-side once and then selectively updated with its own context. (Please close both eyes for the inaccuracies I just wrote 😄 ).

The following composition makes `client-navigation` tests green but breaks `real-world-example` tests since `head` element doesn't update.

```js
<Document>
  <RouterProvider>
    <NavigationProvider>
      <App>
        <Page>
      </App>
    </NavigationProvider>
  </RouterProvider>
</Document>
```

PR #57, opened 2020-12-11
Document rerenders

toomuchdesign:

Bug fix (or at least an experiment)..

#55 

## What is the new behaviour?

Don't re-render Next.js `_document` after initial render (in other words on client-side navigation), and try to rely on `next/head` component side-effects to update `<head>` elements.

Not really, but some edge cases might get a different result.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,32 @@ Next page tester will take care of:
 | **useDocument** (experimental) | Render [Document component][next-docs-custom-document]        | `boolean`          | `false`         |
 | **nextRoot**                   | Absolute path to Next.js root folder                          | `string`           | _auto detected_ |
 
+## Test helpers
+
+Next page tester provides an [utility to setup the expected JSDOM environment](/src/testHelpers.ts).
+
+Run the helper in your global tests setup (in case of Jest It is `setupFilesAfterEnv` file):
+
+```js
+import { initTestHelpers } from 'next-page-tester';
+initTestHelpers();
+```
+
+In case your tests make use of **experimental `useDocument` option**, take the following additional steps:
+
+```js
+import { initTestHelpers } from 'next-page-tester';
+const { setup, teardown } = initTestHelpers();
+
+beforeAll(() => {
+  setup();
+});
+
+afterAll(() => {
+  teardown();
+});
+```
+
 ## Notes
 
 - Data fetching methods' context `req` and `res` objects are mocked with [node-mocks-http][node-mocks-http]
@@ -91,7 +117,7 @@ Note: `document.cookie` does not get cleaned up automatically. You'll have to cl
 
 ### Error: Not implemented: window.scrollTo
 
-Next.js `Link` components invoke `window.scrollTo` on click which is not implemented in JSDOM environment. In order to fix the error you should provide [your own `window.scrollTo` mock](https://qiita.com/akameco/items/0edfdae02507204b24c8).
+Next.js `Link` component invokes `window.scrollTo` on click which is not implemented in JSDOM environment. In order to fix the error you should provide [your own `window.scrollTo` mock](https://qiita.com/akameco/items/0edfdae02507204b24c8).
 
 ### `useDocument` option and `validateDOMNesting(...)` error
 
--- a/README.md
+++ b/README.md
@@ -55,11 +55,18 @@ Next page tester will take care of:
 | **useDocument** (experimental) | Render [Document component][next-docs-custom-document]        | `boolean`          | `false`         |
 | **nextRoot**                   | Absolute path to Next.js root folder                          | `string`           | _auto detected_ |
 
-## Test helpers
+## Set up your test environment
 
-Next page tester provides an [utility to setup the expected JSDOM environment](/src/testHelpers.ts).
+Since Next.js is not designed to run in a JSDOM environment we need to **tweak the default JSDOM environment** to avoid unexpected errors and allow a smoother testing experience:
 
-Run the helper in your global tests setup (in case of Jest It is `setupFilesAfterEnv` file):
+- Provide a `window.scrollTo` mock
+- Provide a `IntersectionObserver` mock
+- Silence `validateDOMNesting(...)` error
+- Remove initial `<head/>` element
+
+Next page tester provides a [helper to setup the expected JSDOM environment](/src/testHelpers.ts) as described.
+
+Run `initTestHelpers` in your global tests setup (in case of Jest It is `setupFilesAfterEnv` file):
 
 ```js
 import { initTestHelpers } from 'next-page-tester';
@@ -86,6 +93,12 @@ afterAll(() => {
 - Data fetching methods' context `req` and `res` objects are mocked with [node-mocks-http][node-mocks-http]
 - Next page tester can be used with any testing framework/library (not tied to Testing library)
 - It might be necessary to install `@types/react-dom` and `@types/webpack` when using Typescript in `strict` mode due to [this bug][next-gh-strict-bug]
+
+### Experimental `useDocument` option
+
+`useDocument` option is partially implemented and might be unstable. `next/head` doesn't currently work on initial render.
+
+`useDocument` might also produce a `UnhandledPromiseRejectionWarning` warning on client side navigation.
 
 ### Next.js versions support
 
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ Since Next.js is not designed to run in a JSDOM environment we need to **tweak t
 - Provide a `IntersectionObserver` mock
 - Silence `validateDOMNesting(...)` error
 - Remove initial `<head/>` element
+- Resolve `next/dist/next-server/lib/side-effect` module in non-brwoser environment
 
 Next page tester provides a [helper to setup the expected JSDOM environment](/src/testHelpers.ts) as described.
 
@@ -96,9 +97,7 @@ afterAll(() => {
 
 ### Experimental `useDocument` option
 
-`useDocument` option is partially implemented and might be unstable. `next/head` doesn't currently work on initial render.
-
-`useDocument` might also produce a `UnhandledPromiseRejectionWarning` warning on client side navigation.
+`useDocument` option is partially implemented and might be unstable. It might produce a `UnhandledPromiseRejectionWarning` warning on client side navigation.
 
 ### Next.js versions support
 
--- a/src/__tests__/use-document/use-document.test.tsx
+++ b/src/__tests__/use-document/use-document.test.tsx
@@ -106,7 +106,7 @@ describe('_document support', () => {
 
   describe('next/head', () => {
     //@NOTE: Missing implementation
-    describe.skip('first render', () => {
+    describe('first render', () => {
       it('merges _document and page head elements', async () => {
         const { page } = await getPage({
           nextRoot: path.join(
--- a/src/__tests__/use-document/use-document.test.tsx
+++ b/src/__tests__/use-document/use-document.test.tsx
@@ -105,7 +105,8 @@ describe('_document support', () => {
   );
 
   describe('next/head', () => {
-    describe('first render', () => {
+    //@NOTE: Missing implementation
+    describe.skip('first render', () => {
       it('merges _document and page head elements', async () => {
         const { page } = await getPage({
           nextRoot: path.join(
--- a/src/__tests__/use-document/use-document.test.tsx
+++ b/src/__tests__/use-document/use-document.test.tsx
@@ -30,12 +30,6 @@ describe('_document support', () => {
         const html = container.querySelector('html') as HTMLHtmlElement;
         expect(html).toHaveAttribute('lang', 'en');
 
-        const metaDescriptions = getMetaTagsContentByName(
-          container,
-          'description'
-        );
-        expect(metaDescriptions[0]).toBe('Custom document description');
-
         const actual = container.querySelector('#__next') as HTMLDivElement;
         actual.removeAttribute('id');
 
@@ -96,16 +90,14 @@ describe('_document support', () => {
         const { container } = render(page);
 
         expect(container.querySelector('head')).toBeInTheDocument();
-        expect(screen.queryByText('Count: 0')).toBeInTheDocument();
-
+        screen.getByText('Count: 0');
         userEvent.click(screen.getByText('Count me!'));
-        expect(screen.getByText('Count: 1')).toBeInTheDocument();
+        screen.getByText('Count: 1');
       });
     }
   );
 
-  describe('next/head', () => {
-    //@NOTE: Missing implementation
+  describe('next/document Head and next/head', () => {
     describe('first render', () => {
       it('merges _document and page head elements', async () => {
         const { page } = await getPage({
--- a/src/__tests__/use-document/use-document.test.tsx
+++ b/src/__tests__/use-document/use-document.test.tsx
@@ -6,6 +6,16 @@ import { getPage } from '../../index';
 import CustomDocumentWithGIP_Page from './__fixtures__/custom-document-with-gip/pages/page';
 import CustomApp from './__fixtures__/custom-document-with-gip/pages/_app';
 
+function getMetaTagsContentByName(element: Element, name: string) {
+  const head = element.querySelector('head') as HTMLHeadElement;
+  const metaTags = head.querySelectorAll(
+    `meta[name="${name}"]`
+  ) as NodeListOf<HTMLMetaElement>;
+  const content: string[] = [];
+  metaTags.forEach((tag) => content.push(tag.content));
+  return content;
+}
+
 describe('_document support', () => {
   describe('_document with getInitialProps', () => {
     describe('with custom _app', () => {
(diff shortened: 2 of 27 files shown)


toomuchdesign:
Since c38b32c, client side navigation works as expected with `useDocument` but now, since `_document` is not re-rendered, it doesn't get any `<head>` update.

I think that rendering `_document` only once is in theory the right way to go, [Next.js docs](https://nextjs.org/docs/advanced-features/custom-document#caveats) give us a few advice:

> - Document is only rendered in the server, event handlers like onClick won't work.
> - React components outside of <Main /> will not be initialized by the browser. Do not add application logic here or custom CSS (like styled-jsx). If you need shared components in all your pages (like a menu or a toolbar), take a look at the App component instead.
> - Document's getInitialProps function is not called during client-side transitions, nor when a page is statically optimized.

It seems to me that the only part of `_document` that needs updates after client side navigation is the `<head>` element. This is the reason why [`next/head`](https://github.com/vercel/next.js/blob/canary/packages/next/next-server/lib/head.tsx#L162) exists, and it seems to update the `<head>` client-side only via context + an [`<Effect>` component](https://github.com/vercel/next.js/blob/canary/packages/next/next-server/lib/side-effect.tsx) which I suppose updating the `<head>` imperatively.

If we could make `next/head` work in JSDOM we're done.

toomuchdesign:
## Support `next/head`

I pushed an extra commit to sound out the feasibility of supporting `next/head`. What I did is to wrap the page with `HeadManagerContext.Provider` when rendered on the client like happens in [`next/client/index.tsx`](https://github.com/vercel/next.js/blob/v10.0.4-canary.4/packages/next/client/index.tsx#L574).

This generates the following error for at least 2 reasons:

```
Warning: next-head-count is missing. https://err.sh/next.js/next-head-count-missing
```

This first reason is that Next.js[ queries the first `head` element in the DOM](https://github.com/vercel/next.js/blob/f4809b950b5df005dafecf92d81509f458668802/packages/next/client/head-manager.ts#L36) while we currently render 2 `head`s because of how `react-testing-library` and `JSDOM` work.

Just as a proof of concept I temporarily replaced the aforementioned line with:

```diff
- const headEl = document.getElementsByTagName('head')[0]
+ const headEls = document.getElementsByTagName('head');
+ const headEl = headEls[headEls.length - 1]
```
With these lines in place the tests get eventually green but we still get a bunch of `next-head-count is missing` errors probably generated by the [`<Effect>` component](https://github.com/vercel/next.js/blob/canary/packages/next/next-server/lib/side-effect.tsx) during unmount phase.

I'll try and figure out a better solution. If we don't manage to find one, we might consider the following temporary options:
- don't support `useDocument` + client navigation at all
- don't support `next/head` updates after initial render

Happy to receive any feedback from you @Meemaw :)

toomuchdesign:
Removing initial JSDOM `<head>` element on Jest's `beforeAll` seems to `next-head-count is missing` from happening. In return an unhandled promise error is now thrown in `real-world-example` tests, but seems something we might be able to work around.

```
(node:73430) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'querySelector' of undefined
```

toomuchdesign:
It turned out that `useDocument` + `<Links>` issue was generated by `<Link>` trying to call global `IntersectionObserver` and then [falling back to `requestIdleCallback`](https://github.com/vercel/next.js/blob/5b89b1b6df1ceed805160c85b3b79771e2ab5ad8/packages/next/client/use-intersection.tsx#L40). Providing a global `IntersectionObserver` stub was enough to avoid the error.

If we merge this PR, we should probably update the README with the list of mocks expected to use `useDocument` option.

I added acceptance tests for `next/head` since there's still a missing point regarding page's missing meta tags at first render. (They work fine after client side navigation)

src/__tests__/use-document/use-document.test.tsx
```diff
@@ -91,4 +103,52 @@ describe('_document support', () => {
...
+        userEvent.click(screen.getByText('Go to A'));
+        await screen.findByText('This is page A');
+
+        const metaDescriptions = getMetaTagsContentByName(
+          container,
+          'description'
+        );
+        expect(metaDescriptions.length).toBe(2);
+        expect(metaDescriptions[0]).toBe('Custom document description');
+        expect(metaDescriptions[1]).toBe('Page A description');
```
Meemaw, line 150: Will there be multiple "descriptions" in NextJS app as well? I think so; just want to make sure.

Meemaw:
Hey @toomuchdesign sorry for late reply. Could you summarise for me status of this PR. Is there anything still not working? I would like to understand tradeoffs (if any) before we merge this and try to exhaust other options.

src/__tests__/use-document/use-document.test.tsx
```diff
@@ -91,4 +103,52 @@ describe('_document support', () => {
       });
     }
   );
+
+  describe('next/head', () => {
+    describe('first render', () => {
+      it('merges _document and page head elements', async () => {
```
Meemaw, line 109: Do we have a test where `_document` + `_app` + `page` Head element?

src/__tests__/use-document/use-document.test.tsx
toomuchdesign: Yep, I was surprised to see that `next/head` simply pushes its meta elements with the ones declared in `_document`.

src/__tests__/use-document/use-document.test.tsx
toomuchdesign: I don't think so, but `next/head` test group in `use-document` might definitely be the right candidate. All that's missing is using `next/head` in `src/__tests__/use-document/__fixtures__/custom-document-with-gip/pages/_app.js`.

This would also make `next/head` tests more complete 👍

toomuchdesign:
Quick recap.

### Broken client side navigation with `useDocument`

In order to not blow up navigation with `useDocument` enabled, `_document` element is only rendered once (initial rendered) and updated imperatively via `next/head`. This is simplistically the way Next.js works. In order to do so I changed the way we compose our elements tree:

```html
<Document>
  <RouterProvider>
    <NavigationProvider>
      <App>
        <Page>
      </App>
    </NavigationProvider>
  </RouterProvider>
</Document>
``` 
In order to get `next/head` updates I had to wrap the page with a "client-side only" `</HeadManagerContext.Provider>`.

### "next-head-count is missing" error

Since Next.js `head-manager.ts` relies on the DOM to have just one `<head>` I had to add an additional `beforeAll` test preparation step to remove initial JSDOM `head`.

We might also consider trying to setup JSDOM manually, to see whether we can get an initial setup more suitable for our special case where we need to render from `html` root element.

### `<Link>` error

I also stumbled upon an error thrown from `<Link>` component (I haven't investigated why it showed up only while testing `next/head`). I fixed it stubbing global `IntersectionObserver`.

My last comment is still valid (I copy-paste it here):

It turned out that `useDocument` + `<Links>` issue was generated by `<Link>` trying to call global `IntersectionObserver` and then [falling back to `requestIdleCallback`](https://github.com/vercel/next.js/blob/5b89b1b6df1ceed805160c85b3b79771e2ab5ad8/packages/next/client/use-intersection.tsx#L40). Providing a global `IntersectionObserver` stub was enough to avoid the error.

If we merge this PR, we should probably update the README with the list of mocks expected to use `useDocument` option.

I added acceptance tests for `next/head` since there's still a missing point regarding page's missing meta tags at first render. (They work fine after client side navigation). This is also missing in current implementation, therefore this PR would provide support for `next/head` during client-side navigation and fix `useDocument` + client-navigation issue.

Meemaw:
Nice recap 👍 

> Quick recap.
> 
> ### Broken client side navigation with `useDocument`
> In order to not blow up navigation with `useDocument` enabled, `_document` element is only rendered once (initial rendered) and updated imperatively via `next/head`. This is simplistically the way Next.js works. In order to do so I changed the way we compose our elements tree:
> 
> ```
> <Document>
>   <RouterProvider>
>     <NavigationProvider>
>       <App>
>         <Page>
>       </App>
>     </NavigationProvider>
>   </RouterProvider>
> </Document>
> ```

Great. Btw, have you tried what happens if we pull `NavigationProvider` to the outermost level and just re-render everything (including document) on navigation? Does it break?

> 
> In order to get `next/head` updates I had to wrap the page with a "client-side only" `</HeadManagerContext.Provider>`.
> 
> ### "next-head-count is missing" error
> Since Next.js `head-manager.ts` relies on the DOM to have just one `<head>` I had to add an additional `beforeAll` test preparation step to remove initial JSDOM `head`.
> 
> We might also consider trying to setup JSDOM manually, to see whether we can get an initial setup more suitable for our special case where we need to render from `html` root element.
> 
> ### `<Link>` error
> I also stumbled upon an error thrown from `<Link>` component (I haven't investigated why it showed up only while testing `next/head`). I fixed it stubbing global `IntersectionObserver`.
> 
> My last comment is still valid (I copy-paste it here):
> 
> It turned out that `useDocument` + `<Links>` issue was generated by `<Link>` trying to call global `IntersectionObserver` and then [falling back to `requestIdleCallback`](https://github.com/vercel/next.js/blob/5b89b1b6df1ceed805160c85b3b79771e2ab5ad8/packages/next/client/use-intersection.tsx#L40). Providing a global `IntersectionObserver` stub was enough to avoid the error.
> 
> If we merge this PR, we should probably update the README with the list of mocks expected to use `useDocument` option.

I'm thinking if we should expose a function that would monkey-patch and quiet all those warning (we have quite a few by now). Users can optionally use it in their `setupTests.ts` or we just point to it from `README.md` so they can re-use its logic.

> 
> I added acceptance tests for `next/head` since there's still a missing point regarding page's missing meta tags at first render. (They work fine after client side navigation). This is also missing in current implementation, therefore this PR would provide support for `next/head` during client-side navigation and fix `useDocument` + client-navigation issue.
> 
> ### P.S.
> No worries about your supposed late reply. It's a pleasure to have you here no matter when :)

toomuchdesign:
Currently `NavigationProvider` relies on `RouterProvider` to listen to route changes. Therefore if I moved the first above I'd have to move the latter, too.

As for tests setup, I'd definitely begin with providing a `jest` setup in the readme for both: with `useDocument` enabled and without. As a second step we might provide both setups in package (even though they might conflict with users' existing setups and of course, they would be tied to Jest).

Meemaw:
> Currently `NavigationProvider` relies on `RouterProvider` to listen to route changes. Therefore if I moved the first above I'd have to move the latter, too.

Makes sense

> 
> As for tests setup, I'd definitely begin with providing a `jest` setup in the readme for both: with `useDocument` enabled and without. As a second step we might provide both setups in package (even though they might conflict with users' existing setups and of course, they would be tied to Jest).

I was thinking something like this:


```js
import { someGoodNameForThis } from 'next-page-tester'

beforeAll(() => {
   someGoodNameForThis() // will monkey patching methods to avoid warnings
})
```

☝️ this doesn't have to be tied to `jest` as we would just monkey patch the methods inside (without using `jest.fn`). It could be useful for starters and examples, but in real world applications it would most likely conflict with user's setup so they could just see the implementation and mock relevant parts themselves.

toomuchdesign:
Added `initTestHelpers`.

README.md
```diff
@@ -55,6 +55,32 @@ Next page tester will take care of:
 | **useDocument** (experimental) | Render [Document component][next-docs-custom-document]        | `boolean`          | `false`         |
 | **nextRoot**                   | Absolute path to Next.js root folder                          | `string`           | _auto detected_ |
 
+## Test helpers
+
+Next page tester provides an [utility to setup the expected JSDOM environment](/src/testHelpers.ts).
```
Meemaw, line 60: I would mention why we are providing this. Something about how `NextJS` uses all those features that are not supported in JSDOM and would just put your console.error on fire 🔥

Meemaw:
Great job @toomuchdesign 🚀 Any ideas why tests are failing?

toomuchdesign:
> Great job @toomuchdesign 🚀 Any ideas why tests are failing?

Yep. I wrote full acceptance tests for `next/head` but we're still not able to render the head elements declared in page at initial render. Only initial render is affected, subsequent client side re-renders are able to update head elements. The same issue effects also master branch.

If we merge we should skip the first `next/head` test and probably add a notice in the docs to point out partial support for `next/head` (even though we already make clear that `useDocument` is an experimental feature)

README.md
toomuchdesign: ✅

Meemaw:
If so, let's just skip those tests for now and create an issue for this and tackle it separately?

I've merged SSR redirects PR so once conflicts resolved I think we're good to go.

toomuchdesign:
Seems we're done here. @Meemaw Would you like to take a look to this PR before we merge?

src/getPage.tsx
```diff
@@ -83,6 +83,7 @@ export default async function getPage({
       return makePage({
         ...mergedOptions,
         route: pageData.redirect.destination,
+        // @NOTE: this option conflicts with `HeadManagerContext` setup (line 97)
         isClientSideNavigation: false,
```
Meemaw: Hmm, I thought I removed this option. It should be removed -- can you do it?

README.md
```diff
@@ -94,6 +94,12 @@ afterAll(() => {
 - Next page tester can be used with any testing framework/library (not tied to Testing library)
 - It might be necessary to install `@types/react-dom` and `@types/webpack` when using Typescript in `strict` mode due to [this bug][next-gh-strict-bug]
 
+### Experimental `useDocument` option
+
+`useDocument` option is not completely implemented and might unstable. `next/head` doesn't currently work on initial render.
```
Meemaw, line 99: `might be unstable`

Meemaw (approved):

toomuchdesign:
I'd like to merge, but CI tests seem to fail way more often than master. I should make sure I haven't introduced any additional issue.

I still haven't figured out what's going on here. Tests never fails locally. Can't tell whether it's an issue with our code or with the way we interact with JSDOM.

Meemaw:
Yeah, it seems this branch is very unstable. Lets get https://github.com/toomuchdesign/next-page-tester/pull/61 in and then rebase and see if it improves.

toomuchdesign:
I've cleaned up this PR to improve readability and to be able to sort of bisect it and find out what's making CI tests unstable.

Meemaw:
I've investigated why head does not work on the initial render (as I was sure this worked before). The problem is that our `evaluateNextJsImportsAsIfOnServer` fires too late and expression in `side-effect` evaluates to false: https://github.com/vercel/next.js/blob/v10.0.3/packages/next/next-server/lib/side-effect.tsx#L3

Moreover, this module evaluates when the

src/__tests__/use-document/use-document.test.tsx
```diff
@@ -6,6 +6,16 @@ import { getPage } from '../../index';
 import CustomDocumentWithGIP_Page from './__fixtures__/custom-document-with-gip/pages/page';
```
Meemaw, line 6: If you remove this import head will work correctly on first render.

toomuchdesign: Good catch about `side-effect.ts`. It could definitely be the cause of initial render problem.

A few days ago I found out that it's possible resetting modules directly from userland code:
https://stackoverflow.com/questions/9210542/node-js-require-cache-possible-to-invalidate/16060619

Could it help us to work around these `isServer` flags?

BTW, with this project I learned that setting `isServer` flags at module level is probably an anti pattern. `typeof window === 'undefined'` should usually run at execution time.

Meemaw: If it's possible to reset modules and we could get that to work that would be the best imo. Otherwise, we cannot control what imports user does in his tests and would need to have a FAQ about this.

toomuchdesign: I'll give it a try.

I won't say this loud, but it seems that `react-hook-form` is the origin of our tests unpredictability. I might be doing something which conflicts with something else in `react-hook-form` which originates a race conditions.

toomuchdesign:
The branch seems stable without `react-hook-form`. I'm investigating whether it could be a behaviour originated by the environment available at module load time. `react-hook-form`, like next.js, stores a `isWeb` variable evaluated at module load time.

src/__tests__/use-document/use-document.test.tsx
toomuchdesign: Jest re-implements `require` and doesn't provide `require.cache` by design.

toomuchdesign:
I eventually:
- required `next/dist/next-server/lib/side-effect` from `initTestHelpers`
- added specific tests targeting data fetching method and global object

`next/head` tests are all green now, and CI seems to be stable again.

README.md
```diff
@@ -55,12 +55,51 @@ Next page tester will take care of:
...
+
 ## Notes
 
 - Data fetching methods' context `req` and `res` objects are mocked with [node-mocks-http][node-mocks-http]
 - Next page tester can be used with any testing framework/library (not tied to Testing library)
 - It might be necessary to install `@types/react-dom` and `@types/webpack` when using Typescript in `strict` mode due to [this bug][next-gh-strict-bug]
 
+### Experimental `useDocument` option
+
+`useDocument` option is partially implemented and might be unstable. `next/head` doesn't currently work on initial render.
```
Meemaw, line 99: It now works on initial render, correct:

src/__tests__/use-document/use-document.test.tsx
```diff
@@ -91,4 +103,53 @@ describe('_document support', () => {
       });
     }
   );
+
+  describe('next/head', () => {
+    //@NOTE: Missing implementation
```
Meemaw, line 108: Is this note still relevant?

toomuchdesign:
We're good to go. 🎉  `useDocument` now works with client navigation and `next/head` is able to update heads' `meta` elements.

`HeadManagerContext.Provider` setup is a bit messy now, but I haven't found an alternative solution that works.

Meemaw:
Lets merge 🚀

Issue #23, opened 2020-11-21
Support server side redirects


Meemaw:
## 🚀 Feature Proposal

We are using SSR with "authenticated" middleware to protect authenticated routes.

This middleware will check if user is authenticated and if not redirect him to another page, e.g. /login

The logic inside `getServerSideProps` might look something like this:

```ts
export const getServerSideProps: GetServerSideProps<Props> = async (
  context
) => {
  if (someSpecialCase) {
      context.res.writeHead(302, { Location: "/login" });
      context.res.end();
      return undefined;
  }
}
```

This means that the I would expect `/login` page to be returned from the `getPage` function.

toomuchdesign:
Hi @Meemaw,
Next 10 introduced [`redirect` and `notFound`  support](https://nextjs.org/blog/next-10#redirect-and-notfound-support-for-getstaticprops--getserversideprops) for `getServerSideProps`.

I'd be inclined toward targeting Next v10 (#20) and adding support for `redirect` and `notFound` result objects. How do you see it?

Meemaw:
Totally agree. That is much easier to implement and a way forward 👍

toomuchdesign:
Currently we already detect result objects lacking `props` field (like `redirect` and `notFound`). next-page-tester will throw [this error](https://github.com/toomuchdesign/next-page-tester/blob/v0.7.0/src/fetchData/fetchPageData.ts#L52) and attach the returned object in `error.payload`.

In some circumstances this might also be good enough depending on the testing approach. Would it fit your testing case?

Since Next.js would trigger an actual server side redirect, we should consider how to translate it into DOM logic carefully since there is currently no way to inform tests about the occurred redirect.

Meemaw:
I would expect not having to catch anything or look into the returned object. Is it not possible that when `redirect` is present -- redirected page is returned?

toomuchdesign:
It should be definitely feasible (even though I still have to find out how). I was considering every option before taking a decision :)

The approach you suggest is definitely closer to an e2e testing experience. Which is interesting.

Meemaw:
Yeah, its very high level and abstract, but I think this is the experience most people would expect and would probably suit most of the use cases. I like the fact that there is no implementation details leaked, you don't even have to know its NextJS underneath.

In theory, library could have multiple "levels" of abstraction and if more fine grained control would be needed some lower level API would be used.

[...]

PR #58, opened 2020-12-13
feat: support server side redirects

Meemaw:

Feature (fixes https://github.com/toomuchdesign/next-page-tester/issues/23)

SSR redirects are not supported

## What is the new behaviour?

SSR redirects are supported

No


Meemaw:
This implements SSR redirects that will automatically follow the `destination` and return final `pageElement`. 

Optionally we could consider adding an option `followRedirects` (with default value of `true`), which would include the `redirect` object in response. (I would never test it like this to be honest; its against the mantra of Testing Library as user does not care about the redirect but cares about the content that he will see on the page).

src/getPage.tsx
```diff
@@ -54,16 +56,32 @@ export default async function getPage({
...
+    const pageData = await fetchRouteData({
+      pageObject,
+      options: mergedOptions,
+    });
+
+    if (pageData.redirect) {
+      return makePage({
+        ...mergedOptions,
+        route: pageData.redirect.destination,
+        isClientSideNavigation: false,
```
toomuchdesign, line 77: Is `pageData.redirect` to be always considered a server-side navigation?

toomuchdesign (changes requested):
Left just 2 comments regarding actual changes.

This PR provides only DOM updates for `redirect` return type, right?

Adding a `followRedirects` option, it could be fine but I'd like to avoid returning different things based on an option value (eg. `page` in case it's `true` and `redirect/pageData` in case of `false`).

I'd try and stay consistent providing always a `page` and optional extra fields.

- What would `redirect` be in case o multiple redirects? Shall we consider returning an array of redirects objects?
- What instead if I got a `redirect` first and a `notFound` object, then? (when we merge this, `notFound` we'll be waiting behind the corner)
- What would be `page` value in case I get a `redirect` with `followRedirects` === `false`? Maybe `<></>`?

I found that being able to return consistent return types is quite useful in case of typescript test files since Testing Library always expect an Element to be rendered. Otherwise TS would complain at every single `render` call.

src/getPage.tsx
Meemaw: Yes, this is only supported on `getServerSideProps` and `getStaticProps` -- both are always executed server side.

package.json
```diff
@@ -39,29 +39,29 @@
...
+    "identity-obj-proxy": "3.0.0",
+    "jest": "26.6.3",
+    "lint-staged": "10.5.3",
+    "next": "10.0.3",
+    "prettier": "2.2.1",
+    "react": "17.0.1",
+    "react-dom": "17.0.1",
+    "react-hook-form": "6.13.0",
+    "ts-jest": "26.4.4",
+    "typescript": "4.1.3"
```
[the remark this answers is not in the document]
Meemaw: I'm up for Renovate. If you set it up it will pin dependencies as well being the most reliable and predictable configuration. See https://docs.renovatebot.com/dependency-pinning/

Meemaw:
> Left just 2 comments regarding actual changes.
> 
> This PR provides only DOM updates for `redirect` return type, right?

Yeah, `redirect` only. Following with `notFound` should be trivial.

> 
> Adding a `followRedirects` option, it could be fine but I'd like to avoid returning different things based on an option value (eg. `page` in case it's `true` and `redirect/pageData` in case of `false`).
> 
> I'd try and stay consistent providing always a `page` and optional extra fields.
> 
> * What would `redirect` be in case o multiple redirects? Shall we consider returning an array of redirects objects?
> * What instead if I got a `redirect` first and a `notFound` object, then? (when we merge this, `notFound` we'll be waiting behind the corner)
> * What would be `page` value in case I get a `redirect` with `followRedirects` === `false`? Maybe `<></>`?
> 
> I found that being able to return consistent return types is quite useful in case of typescript test files since Testing Library always expect an Element to be rendered. Otherwise TS would complain at every single `render` call.

Personally I just brought `followRedirects` up because we were discussing it in the issue https://github.com/toomuchdesign/next-page-tester/issues/23. Personally I don't see much benefits from it and it will only introduce confusion and return type complexity.

package.json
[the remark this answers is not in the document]
toomuchdesign:
```
"rangeStrategy": "bump" :)
```
rangeStrategy

src/getPage.tsx
toomuchdesign: What I referred to is the case of a user navigating client side to a page which triggers a redirect. Next.js doesn't return any 302 but the application performs a seamless client side transition to the final destination route. I was wandering whether this `isClientSideNavigation: false` replicates this behaviour.

toomuchdesign:
Ok, let's keep options simple and just add redirect support. 👍

toomuchdesign (approved):

src/getPage.tsx
Meemaw: Yeah, this already works as expected. I have removed this line as it just made confusion and was wrong (however, it did not influence anything in our logic).

toomuchdesign:
Super. When the time comes, let's have a quick chat about how to implement `notFound` support since I think we might want to figure out what shortcuts to take here.

Issue #64, opened 2020-12-15
Usage with @aws-amplify next/SSR


arelaxend:
Hi there,

## 🚀 Feature Proposal
While using `withSSRContext` from @aws-amplify (https://github.com/aws-amplify/amplify-js) in `getServerSideProps` we get
`ReferenceError: window is not defined`

Should we mock `getServerSideProps` ?

Meemaw:
Hey @arelaxend . `getServerSideProps` is executed on server where `window` is not defined. Does this work for you in Next.JS?

arelaxend:
Sure it does

Meemaw:
Can you tell me which version of `next-page-tester` you are using? After that I can look into what is happening.

Meemaw:
I looked into this and its caused by a very annoying issue we are facing in this library: external modules evaluating `document` | `window` existence on top level (on import).

@toomuchdesign `@aws-amplify` first checks if its browser environment upon import and that evaluates to `true` because of JSDOM, and then tries to access `localStorage` in `getServerSideProps`.

I expect this will cause problems with many external libraries and we probably need a good solution for this.

toomuchdesign:
Hi @Meemaw!
I confirm that `require.cache` (and therefore any library relying on his) is not an option with `jest`.

If we could split our initial rendering process in 2 consecutive phases (server, browser) we might consider:
- temporarily remove browser global objects (on tests setup)
- execute server logic
- restore browser global objects
- resetting modules

The last step would be framework-specific. We might consider:
- detecting Jest (`typeof jest !== 'undefined'` 😄 )
- clearing modules via `jest.resetModules()`
- clearing modules via `require.cache` in case no known test frameworks are detected
- optionally provide a module clearing callback configurable by users

I suspect this would still leave room for bugs, though.

[This](https://github.com/toomuchdesign/next-page-tester/blob/master/src/getPage.tsx#L118) could be a good line to reset modules and browser global objects.

Another option could involve using Jest in `node` environment, and initialise JSDOM on our side before we return page element.

How about adding some failing tests [here](https://github.com/toomuchdesign/next-page-tester/tree/master/src/__tests__/global-object) so that we can reproduce the problem and experiment solutions?

toomuchdesign:
A fix is available in #73 which is blocked until Jest v27 gets released.

PR #67, opened 2020-12-17
test: suite for expected global object behaviour

Meemaw:

## What is the new behaviour?

...

### DIFF
--- a/src/__tests__/global-object/__fixtures__/pages/gip.js
+++ b/src/__tests__/global-object/__fixtures__/pages/gip.js
@@ -1,17 +1,28 @@
 import React from 'react';
 import { PropsPrinter } from '../../../__utils__';
+
 const window_moduleLoadTime = typeof window !== 'undefined';
 const document_moduleLoadTime = typeof document !== 'undefined';
+const isWeb_moduleLoadTime = window_moduleLoadTime && document_moduleLoadTime;
 
 export default function gip(props) {
+  const window_componentScope = typeof window !== 'undefined';
+  const document_componentScope = typeof document !== 'undefined';
+  const isWeb_componentScope = window_componentScope && document_componentScope;
+
   return (
     <>
       <h2>Page</h2>
       <PropsPrinter
         props={{
           ...props,
-          window_componentScope: typeof window !== 'undefined',
-          document_componentScope: typeof document !== 'undefined',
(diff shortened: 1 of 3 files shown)


Meemaw:
@toomuchdesign added some more tests for the problematic behaviour regarding module imports (see https://github.com/toomuchdesign/next-page-tester/issues/64). I think if we are somehow able to make these tests pass, it will solve all problems.

toomuchdesign:
`jest.isolateModules` has a few open issues. Linking here just for documentation purpose:

- [jest.isolateModules not accepting async functions](https://github.com/facebook/jest/issues/10428)
- [does not work if module has already been required with a non-isolated method](https://github.com/facebook/jest/issues/7863)

toomuchdesign:
Brief report after a few attempts of working around ES/`jest` modules and global objects.

`jest.isolateModules` seems quite a powerful pattern but it currently comes with a couple of unexpected limitations.

### jest.isolateModules relativity

I confirm that [this issue](https://github.com/facebook/jest/issues/7863) still occurs and `jest.isolateModules` doesn't load a fresh module if that modules is already imported outside a `jest.isolateModules` call.

This means that, in order to ensure we get modules isolation, we should wrap all tests in a `jest.isolateModules` call :)

### jest.isolateModules + Typescript

Since we have to use `require` from inside `jest.isolateModules`, I found out that Typescript looses typing information of those modules imported with `require`. Maybe I'm missing a TS config or something else? Eg:

```js
jest.isolateModules(() => {
  const { createElement } = require('react');
  // createElement: any
});
```

### React multiple instances error

Taking the opposite approach with `jest.resetModules()` took me to the error where React detects hook calls to different React instances.

```
Error: Uncaught [Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
```

Meemaw:
Didnt dive into the problem yet, but there might be some ideas in this issue: https://github.com/facebook/jest/issues/8987

toomuchdesign:
Well, it seems we're going to have `jest.isolateModules` fixed: https://github.com/facebook/jest/pull/10963 😄

PR #71, opened 2020-12-20
chore: log un-awaited navigation routes

Meemaw:

Improvement

Warning is logged on un-awaited client side navigation, but it lacks some additional information to make it easier to debug which transition the warning relates to (its hard to pinpoint the problem when jest is executing tests in parallel and logging all over the place)

## What is the new behaviour?

Previous and next route are included in the log

No


toomuchdesign (approved):
👍

PR #74, opened 2020-12-24
test: aws-amplify integration

Meemaw:
Test case for https://github.com/toomuchdesign/next-page-tester/issues/64

(this fails on master)


toomuchdesign (approved):
All good, thanks @Meemaw! I expected error described in #64 to show up, but I'm probably missing context.

Are we migrating `real-world-test` folder as a `third-party-integration` test case in a separate PR?

PS. What if we shortened `third-party-integration` as `third-party`? Shorter to refer to but maybe less explicit? I'll leave it to you :)

Meemaw:
> All good, thanks @Meemaw! I expected error described in #64 to show up, but I'm probably missing context.
> 

Its targeting your branch (`server-pages`) thus it works as expected. You would see error described in #64 on master though.

> Are we migrating `real-world-test` folder as a `third-party-integration` test case in a separate PR?

Yes, I would to it after all of this is merged.

> 
> PS. What if we shortened `third-party-integration` as `third-party`? Shorter to refer to but maybe less explicit? I'll leave it to you :)

Done

toomuchdesign:
I missed your initial comment! Let's park it then, until Jest 27 gets released.

PR #76, opened 2020-12-28
test: baseweb integration

Meemaw:

Bug fix: correctly render `_app` in server context when `useDocument=true`

## What is the new behaviour?

...


Meemaw:
@toomuchdesign I've prepared a test case where stuff seems to break internally. After short investigation it seems that `client` version of custom `_app` is not loaded correctly; resolves to `{}` and `DefaultApp` is used. I'm not entirely sure why this happens though, any ideas?

In some cases this might happen if there is a circular dependency on the imported module but I don't see that being the case here.

src/__tests__/third-party/baseweb/pages/_app.tsx
```diff
@@ -0,0 +1,19 @@
...
+import App from 'next/app';
+import { Provider as StyletronProvider } from 'styletron-react';
+import { LightTheme, BaseProvider } from 'baseui';
+import { styletron } from '../styletron';
+
+export default class MyApp extends App {
+  render() {
+    const { Component, pageProps } = this.props;
+    return (
+      <StyletronProvider value={styletron}>
```
Meemaw, line 11: The problem seems to happen when wrapped with `<StyletronProvider value={styletron}>`

src/__tests__/third-party/baseweb/pages/_document.tsx
```diff
@@ -0,0 +1,51 @@
...
+  DocumentContext,
+  Head,
+  Html,
+  Main,
+  NextScript,
+} from 'next/document';
+import { Provider as StyletronProvider } from 'styletron-react';
+import { Server, Sheet } from 'styletron-engine-atomic';
+import type { RenderPageResult } from 'next/dist/next-server/lib/utils';
+import { styletron } from '../styletron';
```
Meemaw, line 12: For some reason `styletron` is undefined here 🤔

src/__tests__/third-party/baseweb/pages/_document.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export default class CustomDocument extends Document<Props> {
+  static async getInitialProps(ctx: DocumentContext) {
+    const page = ctx.renderPage((App) => (props) => (
+      <StyletronProvider value={styletron}>
+        <App {...props} />
+      </StyletronProvider>
+    ));
+
+    const stylesheets = (styletron as Server).getStylesheets() || [];
```
Meemaw, line 26: For some reason `styletron` is undefined here and this crashes 💥

toomuchdesign: One potential issue might come from the fact that we render the [custom document in `client` mode](https://github.com/toomuchdesign/next-page-tester/blob/3512c8fd2f97cce413181fbdb4d82bba5d856c7f/src/_document/render.tsx#L26) since the server version doesn't currently work. It's the opposite way Next.js deals with document component: rendered once on the server as string and never rehydrated.

I see that `styletron` file makes some environment detection, but only the client version is currently used, since `next-page-tester` renders custom document in client environment (while Next.js does the opposite):
https://github.com/toomuchdesign/next-page-tester/blob/b1e4df6ef251d7a8d15cc45bb145ef40cce2f477/src/__tests__/third-party/baseweb/styletron.ts#L9

The best solution would consist of rendering custom document as Next.js does. More or less like:
- Render document element in server environement
- Output it as string
- Remove the contents of root div `#__next` element
- Mount app component as a child of `#__next` element

This would also maybe fix the issue described here, but I haven't been able to get there at the moment.

If this is the problem, could it be temporarily addressed in `baseweb/styletron.ts`?

Meemaw: I kept that file as it is intentionally. Almost everyone using that library will use it like this. It's from a boilerplate. Point being, users should not be forced to change their code to use this library.

src/testHelpers.ts
```diff
@@ -40,11 +40,16 @@ export function initTestHelpers() {
 
   function setup() {
     if (isJSDOMEnvironment()) {
-      // Remove initial JSDOM <head> element
-      const headElement = document.querySelector('head');
-      if (headElement) {
-        headElement.remove();
```
Meemaw, line 46: This was the reason that `_app` import did not resolve correctly. At import time construction of `styletron's Client` failed because of this line: https://github.com/styletron/styletron/blob/master/packages/styletron-engine-atomic/src/client/client.js#L209

src/utils.ts
```diff
@@ -115,6 +115,7 @@ export function executeWithFreshModules<T>(f: () => T): T {
       // Ensure every page gets the same patched 'next/document' instance
       // imported before tests by "src/testHelpers.ts"
       jest.mock('next/document', () => jest.requireActual('next/document'));
+      jest.mock('react', () => jest.requireActual('react'));
```
Meemaw, line 118: Any ideas why this was not needed before and started popping up now? Anyhow I think we are safer if we have it explicitly required.

Meemaw:
@toomuchdesign got this to work 🙂 Let me know what you think.

src/_app/render.tsx
```diff
@@ -14,12 +14,18 @@ export default function renderApp({
...
+  if (useApp) {
+    if (options.useDocument) {
+      if (customAppFile?.server.default) {
+        AppComponent = customAppFile.server.default;
+      }
+    } else {
+      if (customAppFile?.client.default) {
+        AppComponent = customAppFile.client.default;
+      }
+    }
```
toomuchdesign, line 28: Are you sure we're using the proper `customAppFile` here? It seems to me that the `customAppFile` returned to the user at initial render should always be the one initialised in `client` environment (no matter what `useDocument` value).

This bring us back to the current -I'd say- architectural limitation of `next-page-tester`: we currently never perform a full tree render in `server` mode (like Next.js obviously does during SSR).

We directly (and roughly) return an element tree similar to the one generated in-browser after React hydration. This solution works fine as long as we consider `_app` and pages.

We're currently leaving out this complexity and we might have to find out whether we can work around this limitations or we should provide one or two different ways of rendering with `next-page-tester`.

Out of topic: we might be technically able to return from `getPage` with:
- a `html` string representing the SSR-rendered page to inject in JSDOM
- an elements' tree to mount with (React.hydrate) on the `#__next` div

This would definitely be closer to how I understand Next.js works (ad also a promising approach), but we should find out how to get a good dev experience out of it.

src/testHelpers.ts
```diff
@@ -40,11 +40,16 @@ export function initTestHelpers() {
...
-      }
+      // This is needed by NextJS to correctly find the head element which is rendered in `document.body`
+      // https://github.com/vercel/next.js/blob/c8cd77a856248346768042f4553a8513195311a4/packages/next/client/head-manager.ts#L36
+      const getElementsByTagName = document.getElementsByTagName;
```
toomuchdesign, line 45: I'm not fully into the context of this, but I expect a lot of troubles coming from patching global methods.

Could it be an alert the we're doing something wrong? (Let's forget for a moment that rendering a `Next.js` app in JSDOM is already a wrong idea per se 😄 ).

src/utils.ts
toomuchdesign: I don't know why it popped up just now, but I felt we'd have been forced to do it sooner or later.

I'm realising that since we rely on this, users should never reset their modules, right?

toomuchdesign (changes requested):
Hi @Meemaw, thanks for this great testing efforts. Really appreciated :)

A left a few comments which might cover a wider spectrum then this specific PR, but it seemed the right context to explore them.

src/_app/render.tsx
Meemaw: I agree, we should be returning the `client` customApp file to the user. Problem is that the client version fails at the `renderToStaticMarkup` in this specific use-case.

I think its a compromise at this point:
1. Have `_app` correctly render server side (I think this means client side events wont work in `_app` if you had any -- e.g `onClick` handlers or `Link` -- I have never used this and would expect is not so frequent)
1. Rendering `_app` with wont work for some projects and will crash

As you mentioned, we are already hacking a lot of stuff and there will surely be limitations/tradeoffs with the current implementation.

With this (and other) PRs I just wanted to focus on more realistic examples of NextJS applications so we have those covered as much as possible. While everything is perfect for simple "todo apps", I'm afraid it still doesn't work for a lot of non-trivial projects out there.

src/testHelpers.ts
Meemaw: I think we are doing best we can with what we have (JSDOM) 😄 Not sure if you saw my comment why this was needed: https://github.com/toomuchdesign/next-page-tester/pull/76/files#r550180943

Meemaw:
@toomuchdesign I'm not entirely convinced if we should merge this and insisting with the current "approach" or try to investigate the "hydration" alternative.

toomuchdesign:
From the maintainer perspective, I'd be generally inclined to publish something which has (documented) limitations and does what it does as right as possible.

I'm afraid that introducing a deviation from the expected behaviour to support a specific user-scenario might make harder documenting this and further development.

I don't know f it's the case, but in general I'd prefer to ask users to mock 1 dependency on their side instead of introducing a possible layer of complexity and failure here. This even more vital considered the limited resources we have as maintainers.

Btw, I'll take some time to better focus the issue and land a hand if I'm able to. Maybe there are different approaches that could work. Let's see. 🚀  ☃️

src/testHelpers.ts
toomuchdesign: Now I got full context about this part. The trade-off is: always render an initial document with missing `head` vs. patching global `document.getElementsByTagName` to tweak the way Next.js retrieve pages' `head` element. 🤔

toomuchdesign:
I'd like to give the SSR render + client side hydration approach a try. When I have time I'll try to open an explorative PR. I'd try to not focus on DX and keep tests as much untouched as possible. So that we can have an idea of the possible blockers.

In the meanwhile we could release what we have in master. How do you see it?

Meemaw:
Agree, lets release 🚀

PR #79, opened 2020-12-29
feat: execute and require app component in browser and non-browser environment

toomuchdesign:

Bug fix / feature

1- Pages receive `pageProps` from custom app even when `useApp` is set to `false`
2- App component's fetching methods are always executed in browser context

## What is the new behaviour?

1- Ensure `useApp = false` prevents custom app from fetching data whatsoever
2- Execute (and import) App `getInitialProps` method in browser or non-browser context based on application lifecycle (initial render or client side navigation)

Yes.


Meemaw (commented):
I actually never used data fetching methods in `_app` but always though `_app` is a "client" side thing. Are the data fetching method executed server side in Next?

toomuchdesign:
At initial render I'd expect `_app` to execute `gip` and itself server side and returned as HTML. Just like pages.

Meemaw (approved):

PR #82, opened 2020-12-31
[WIP] Render html server side and hydrate in JSDOM

toomuchdesign:

Refactor

`next-page-tester` has an important architectural limitation: it currently never performs a full tree render in server mode (like Next.js obviously does during SSR).

This process is skipped in favour of directly (and roughly) return an element tree similar to the one generated in-browser after React hydration. This solution works fine as long as we consider `_app` and pages.

## What is the new behaviour?

Here we explore the possibility of changing the way `next-page-tester` renders the element tree and what it returns to the user. The main changes would affect initial cycle (and optimistically leave client side navigation untouched).

Initial render could happen like this:

- Fetch route data as usual
- Create 2 component trees of the whole page: one emulating the way it should be generated server side, and one client side.

`getPage` would return:
- a `html` string representing the SSR-rendered page element (to be inject in JSDOM)
- the usual `page` elements' tree to mount (with React.hydrate) on `#__next` div

😄😄😄😄😄😄😄 

## Other information:

@Meemaw feel free to comment, branch and do whatever makes you feel creative.


src/getPage.tsx
```diff
@@ -93,15 +119,21 @@ export default async function getPage({
...
-      pageElement,
+    serverPageElement = await renderDocument({
+      pageElement: serverPageElement,
       options,
       pageObject,
       pageData,
     });
   }
 
+  const html: string = ReactDOMServer.renderToStaticMarkup(
```
Meemaw, line 130: We already `renderToStaticMarkup` in here: https://github.com/toomuchdesign/next-page-tester/blob/master/src/_document/render.tsx#L45. Cant we reuse that? 

Moreover, I think if we make this work flawlessly, I don't see a reason why `useDocument` and `useApp` would stay configuration options. They just complicate the code and increase the API surface while perhaps keeping users confused (as they would expect the whole page to render)?

toomuchdesign: 1- If we always rendered a `_document` we could always end up rendering the tree as string with `renderDocument`. Do we ever call that `renderPage` method or it's something we provide to Next.js internals? Maybe we could be able to end up with an implementation closer to Next.js, since now we should only render `_document` as a string, right?

2- I'd say yes, even though ignoring `_app` and _document` rendering in complex applications could help leaving unnecessary mocks/setups out of tests (?).

If it worked it would also allow to get rid of most of our `testHelpers.ts` setup. Let's see how it evolves.

toomuchdesign:
The first brutal hydration test works. :)

Meemaw:
Adding some comments about the API;

It seems with current implementation we are coupling the library with `@testing-library/react`. I think we can avoid that and still have a nice API.

```js
const { html, pageElement, container } = await getPage({ route: '/' })
rtl.hydrate(pageElement, container)

expect(screen.getByText('Hello world!')).toBeInTheDocument();
```

Note that for this we would have to set `document.body.innerHtml` inside `getPage` and query it to retrieve `container` (pointing to the `__next` div).


I haven't really tried that in branch locally if it works, but its something I think would work nicely for me as a user of the library.

toomuchdesign:
Hi @Meemaw,
I had the same goal in mind about decoupling the library from `testing-library`. I used `testing-library` in render method just as an intermediate step, but I'd definitely replace it with `react-dom`'s `hydrate`.

Actually final users should not even need `react-testing-library` or `enzyme` but simply use a plain DOM testing utility (like `testing-library`), since we all the reactish logic stays inside `next-page-tester` boundaries.

I'm currently stuck with the usual "multiple react instances error". The error showed up at commit 86a1228 when I refactored `_app` logic to load `client` and `server` versions of App component and behave accordingly to current `env`. I didn't really expect it to happen again :)

I'd be happy if you could take a look at it. (I temporarily disabled `useDocument` tests to decrease tests failures noise). Maybe there's something about Jest  modules that I'm still missing. I'll try to investigate.

If we were able to overcome this issue, this is what comes next:
- refactor `_document` logic to initially render as string
- enable `global-object` tests and make sure they assert what they should
- refactor `render` method (and tests) to not use `testing-library`

Meemaw:
One idea: perhaps it would be smart to create a new branch (e.g. `v1`) from master and try to split this into multiple PRs targeting it so its easier to isolate such problems.

I think refactors (e.g. loading `_app` according to `env` and such) could be merged without breaking stuff. I also did that in https://github.com/toomuchdesign/next-page-tester/pull/76 and tests are passing.

toomuchdesign:
Good idea @Meemaw.

I moved the initial commits to `next` branch and opened 2 PR's marked with [next] prefix: one for `render` + `renderHtml` methods, one for rendering `_app` in server env.

toomuchdesign:
Superseded by #90

PR #84, opened 2021-01-02
[next] feature: render app based on env

toomuchdesign:

Feature

App always renders as client environment

## What is the new behaviour?

Render as server environment first, and then as client at mount and client side navigation.

Yes. App render result might differ from previous one.

## Other information:

~Rendering `_App` in server mode brings back the "multiple React instances" error. In `src/_app/render.tsx` I marked the lines which generates the error and the previous working implementation.~

There are not currently tests covering this change because `getPage` doesn't return the output rendered in `server` mode. We might add them (`global-object` tests) once #83 gets merged.


src/_app/render.tsx
```diff
@@ -11,21 +10,31 @@ export default function renderApp({
...
-  if (useApp && customAppFile?.client?.default) {
-    AppComponent = customAppFile.client.default;
-  } else {
-    AppComponent = DefaultApp;
-  }
-
-  return (
+  // @NOTE This generate "multiple React instances" error
+  return env === 'server' ? (
+    executeAsIfOnServerSync(() => (
```
Meemaw, line 20: I think we don't need to wrap it with `executeAsIfOnServerSync` here and I assume that is causing the problems. We are not rendering here, but just wrapping JSX.Element, which is of correct environment already.

src/utils.ts
```diff
@@ -114,7 +113,12 @@ export function executeWithFreshModules<T>(f: () => T): T {
     jest.isolateModules(() => {
       // Ensure every page gets the same patched 'next/document' instance
       // imported before tests by "src/testHelpers.ts"
-      jest.mock('next/document', () => jest.requireActual('next/document'));
+      jest.doMock('react', () => jest.requireActual('react'));
```
Meemaw, line 116: As this list is increasing, I think we could extract it to some list to always be sure they are in sync with the `stealthyRequire`.

```js
const nonIsolatedModules = ['react, 'react-dom']


nonIsolatedModules.forEach((nonIsolatedModule) => {
   jest.doMock(nonIsolatedModule, () => jest.requireActual(nonIsolatedModule));
})

nonIsolatedModules.forEach((nonIsolatedModule) => {
   require(nonIsolatedModule)
})
 ```

File: src/_app/render.tsx [hunk 3]
```diff
@@ -11,21 +10,31 @@ export default function renderApp({
...
-      Component={pageObject.page.client.default}
+      Component={pageObject.page[env].default}
       pageProps={pageData.props}
     />
   );
+
+  // @NOTE This works but doesn't complain to client/server env expectations
+  // return (
+  //   <AppComponent
+  //     Component={pageObject.page.client.default}
```
Meemaw, line 36: Related to the comment above, I think we can just use:

```js
return (
   <AppComponent
        Component={pageObject.page[env].default}
        pageProps={pageData.props}
   />
```

src/_app/render.tsx
toomuchdesign: If I get the whole picture, we need `executeAsIfOnServerSync` here to ensure that components are executed (not loaded) in `server` environment (eg. getting the right result from `typeof window === 'undefined'`).

I found out that `jest.requireActual` works as expected when the module to be requires gets required before the `jest.requireActual` call 🤔 . Maybe another Jest bug?

src/utils.ts
toomuchdesign: I definitely overdid with these non isolated dependencies. `react` and `next/document` were enough. The issue was `jest.requireActual` expecting the module being imported beforehand.

toomuchdesign:
Now  "multiple React instances" error is over. There are currently troubles with components doing `useRouter` calls, but there's room to investigate, now.

src/_app/render.tsx
toomuchdesign: BTW once we add tests for this change we'll be able to dispel any doubt.

Meemaw: By execution do you mean being rendered? So we have 3 lifecycle "phases" of each "component":
- import time ✅ 
- data fetching time ✅ 
- render time (this happens after -- when we are rendering the component -- `renderToStaticMarkup`  or `rtl.render`)

I might be wrong, but at this point we are just wrapping the component so I think its redundant. Anyhow, don't we have tests for global `_app` that would confirm my suspision?

toomuchdesign:
Surprisingly enough (😄) it was a matter of modules imported as fresh instances which probably broke React context providers: `next/router`, `next/head`, `next/dist/next-server/lib/router-context`.

Maybe in a further refactor we might try to resolve/execute the whole `server` logic inside a single isolated `jest.isolatedModules` call and avoid some of this non-isolated modules setup.

src/_app/render.tsx
toomuchdesign: I guess you're right @Meemaw :)

toomuchdesign: We have tests global-object tests, but we're not testing again SSR result because the current API doesn't expose SSRed html.

Meemaw (approved):

src/testHelpers.ts
```diff
@@ -1,5 +1,7 @@
 import { executeAsIfOnServerSync } from './server';
 
+require('react');
+require('react-dom');
```
Meemaw, line 4: Are these needed?

toomuchdesign: No!

Meemaw:
I think this one could target `master` as it is actually a bug fix.

toomuchdesign:
We probably could merge it to master without breaking tests, even though currently master never renders in SSR mode.

With this new approach we're actually adding a 3rd lifecycle: before it was `initial client render` + `client navigation` now we add a `SSR render` pass which is only available as string and doesn't even exists as React's runtime.

I'm realising that this might also help spotting possible differences between SSR and client output directly in unit tests.

PR #85, opened 2021-01-03
fix: add req.cookies property in "getServerSideProps"

Meemaw:

Bug fix -- add parsed `cookies` property to `GetServerSidePropsContext`. This is available from Next `10.0.4` and was added in this PR: https://github.com/vercel/next.js/pull/19724

`req.cookies` are empty in `GetServerSidePropsContext`

No

### DIFF
--- a/src/fetchData/makeHttpObjects.ts
+++ b/src/fetchData/makeHttpObjects.ts
@@ -1,17 +1,16 @@
 import httpMocks from 'node-mocks-http';
 import type { OptionsWithDefaults, PageObject } from '../commonTypes';
+import { parse } from 'cookie';
 
 export default function makeHttpObjects({
   pageObject: { params, route },
   reqMocker,
   resMocker,
-  appendCookie,
   refererRoute,
 }: {
   pageObject: PageObject;
   reqMocker: OptionsWithDefaults['req'];
   resMocker: OptionsWithDefaults['res'];
-  appendCookie?: boolean;
   refererRoute?: string;
 }) {
   const req = httpMocks.createRequest({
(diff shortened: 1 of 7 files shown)


src/server.ts
```diff
@@ -15,12 +15,12 @@ export const executeAsIfOnServer = async <T>(f: () => T) => {
...
-
-  global.window = tmpWindow;
-  global.document = tmpDocument;
-
-  return result;
+  try {
+    return await f();
+  } finally {
+    global.window = tmpWindow;
+    global.document = tmpDocument;
```
Meemaw, line 22: Not really related to this fix, but I noticed this is an issue when experimenting with custom `_error` page. When an `Error` is thrown from some data fetching method we didn't restore `client` environment which lead to crash at render time, because `document`  was not defined.

src/fetchData/makeHttpObjects.ts
```diff
@@ -1,27 +1,24 @@
...
   refererRoute?: string;
 }) {
   const req = httpMocks.createRequest({
     url: route,
     params: { ...params },
   });
 
-  // Make document.cookie available in req.headers
-  // @NOTE: SHall we make available req.headers, too?
-  if (appendCookie && document && document.cookie) {
```
Meemaw, line 24: `appendCookie` was always `true` thus redundant.

src/server.ts
toomuchdesign: I had the impression something like this was happening yesterday when I had several broken tests. Clever idea 👍

toomuchdesign (approved):
Seems very good to me. Thank you @Meemaw!

PR #86, opened 2021-01-03
[next] test migration

toomuchdesign:

Test update

Current tests retrieve `page` element from `getPage` return object and render it with `@testinbn-library/react`

## What is the new behaviour?

Since we're considering switching to a new render strategy based on pre-render SSR html in JSDOM and then mount `page` element into it with a `hydrate` call (using `render` or `renderHTML` methods), tests should be rewritten to adhere to the new testing flow.

No, but it's part of a wider breaking change.

## Other information:

The migration is still ongoing...

### DIFF
--- a/src/makeRenderMethods.ts
+++ b/src/makeRenderMethods.ts
@@ -24,6 +24,8 @@ export default function makeRenderMethods({
     document.documentElement.replaceChild(originalBody, document.body);
 
     const nextRoot = document.getElementById('__next');
+
+    /* istanbul ignore next */
     if (!nextRoot) {
       throw new Error('[next-page-tester] Missing __next div');
     }
--- a/pages/page.js
+++ b/pages/page.js
@@ -1,5 +1,5 @@
 import React from 'react';
 
 export default function PagesDiscoveryInRoot() {
-  return <>/auto-pages-directory/in-root/pages/page</>;
+  return <>Page in natural root</>;
 }
--- a/src/__tests__/__utils__/PropsPrinter.tsx
+++ b/src/__tests__/__utils__/PropsPrinter.tsx
@@ -1,6 +1,16 @@
 import React from 'react';
 import { stringify } from './index';
 
-export function PropsPrinter<T extends object>({ props }: { props: T }) {
-  return <>`props: ${stringify(props)}`</>;
+export function PropsPrinter<T extends object>({
+  suppressHydrationWarning,
+  ...props
+}: {
+  suppressHydrationWarning?: boolean;
+  props: T;
+}) {
+  return (
+    <div suppressHydrationWarning={suppressHydrationWarning}>
+      `props: ${stringify(props)}`
+    </div>
+  );
 }
--- a/src/__tests__/__utils__/expectDOMElementsToMatch.ts
+++ b/src/__tests__/__utils__/expectDOMElementsToMatch.ts
@@ -1,17 +1,24 @@
 const REACT_NEW_LINE_COMMENTS_REGEX = /<!-- -->/g;
 const REACT_DATA_REACT_ROOT = / data-reactroot=\"\"/g;
-export function stripReactReactExtraMarkup(string: string): string {
+export function stripReactExtraMarkup(string: string): string {
   return string
     .replace(REACT_NEW_LINE_COMMENTS_REGEX, '')
     .replace(REACT_DATA_REACT_ROOT, '');
 }
 
+/*
+ * Since initial JSDOM dom tree is generated from a html string
+ * created with "ReactDOMServer.renderToString",
+ * dom elements have some extra react-specific extra markup which cause
+ * plain deepEquality assertions to fail.
+ * Here we strip the extra markup before the equality assertion.
+ */
 export function expectDOMElementsToMatch(
   actual: Element,
   expected: Element
 ): void {
-  const actualString = stripReactReactExtraMarkup(actual.outerHTML);
-  const expectedString = stripReactReactExtraMarkup(expected.outerHTML);
+  const actualString = stripReactExtraMarkup(actual.outerHTML);
+  const expectedString = stripReactExtraMarkup(expected.outerHTML);
 
   return expect(actualString).toEqual(expectedString);
 }
--- a/src/__tests__/__utils__/index.ts
+++ b/src/__tests__/__utils__/index.ts
@@ -5,3 +5,5 @@ export { sleep } from './sleep';
 export { stringify } from './stringify';
 export { wrapWithNextRoot } from './wrapWithNextRoot';
 export { expectDOMElementsToMatch } from './expectDOMElementsToMatch';
+export { makeNextRootElement } from './makeNextRootElement';
+export { renderWithinNextRoot } from './renderWithinNextRoot';
--- a/src/__tests__/__utils__/makeNextRootElement.ts
+++ b/src/__tests__/__utils__/makeNextRootElement.ts
@@ -1,0 +1,5 @@
+export function makeNextRootElement(): Element {
+  const root = document.createElement('div');
+  root.id = '__next';
+  return root;
+}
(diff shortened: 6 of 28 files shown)


toomuchdesign:
Hi @Meemaw,
I quickly setup this PR to start to have a feeling of what kind of testing experience we want to provide with the new API's and rendering flow. I'd love to hear your thoughts.

Here I migrated a first chunk of tests: router, router-mocking, ssr-redirect and use-app. 

I think we're definitely improving the overall user testing quality compared to what we have now, but there are also a few details which could be ideally improved.

### `render` return type
Back to your suggestions, the mounted root `__next` element is currently not provided by `getPage` (since at that time the DOM is still untouched), but by the `render` methods returned by `getPage`. 

-1 How do you see it? Shall we consider returning an object from `render` instead of the plain root element?

### `render` return type II
As you can see, in our tests, we often rendered a page with `getPage` and then manually render the expected page component to compare to each other.

Since  `const {render} = getPage()` returns the actual page wrapped with `__next` root, this means that we need to also wrap the expected page element (the one we manually render) with a `__next` element, making the testing flow a bit clunkier then before.

I expect most of users not to test like this but to assert against existing DOM, but it's worth mentioning this detail.

Do you think we could do better?

### Comparing SSRed DOM elements with other elements
The aforementioned testing strategy (nested elements deep equality), is also complicated by the fact that we render initial DOM from the string returned from `ReactDOMServer.renderToString`. The generated string has some extra react-specific markup (especially html comments) which cause deepEquality comparison to fail.

I temporarily worked around this issue with `src/__tests__/__utils__/expectDOMElementsToMatch.ts` (we might also try and setup a custom expect assertion). This would also affect real users testing flow (even though most of them should simply assert against existing mounted DOM).

A can't see any possible improvement on this side, even because we're replicating what React and Next do. After all it's a kind of complexity which brings closer to what we're trying to replicate.

Feel free to express any concern, of course.

Well, That's all for now. I'll complete test migration in the next days and I'm looking forward to hear your thought.

PS. I still have to see how to render `document` component with the new rendering strategy, I'll might need your support there :)


Thank for you priceless help @Meemaw 🙌

Meemaw:
Hey @toomuchdesign 

I'm still wondering if we should just "pre-render" in the `getPage` function and among others, return the `__next`  container. This would allow following patterns:

// Without RTL
```js
const { render } = await getPage({ route })
render();
```

// With RTL hydration
```js
const { page, container } = await getPage({ route })

RTL.hydrate(page, container)

await screen.findByText("Hello world")
```

// Without RTL hydration (probably should be discouraged, unless user does not care about `_document`)
```js
const { page } = await getPage({ route })

RTL.render(page)

await screen.findByText("Hello world")
```

☝️ note that this would be a non-breaking change. Last options still looks the same as the current approach.

toomuchdesign:
Hi @Meemaw 👋,
if I get your examples properly, **Without RTL** and **Without RTL hydration** are currently already supported.

As for "With RTL hydration" we might easily get there if current `renderHTML` method returned the `__next` root element generated at first SSR pass (as `render` already does):

```js
const { page, renderHTML } = await getPage({ route })
const container = renderHTML()
RTL.hydrate(page, container)

await screen.findByText("Hello world")
```
I'll throw a commit for it 👍 

Regarding your last point - **Custom hydrate option** -  it's something I'm also reasoning about. Even though I found out that with a single `within` call we can leverage all the benefits of `@testing-library/dom` (after all global `screen` would be equivalent to `const screen = within(container)`), I'm open to consider a shortcut for it.

One solution might consist of giving users the ability of redeclaring their `render` method implementation (we should only find a way to do it globally)

Another one might consist of somehow providing `@testing-library` bindings for `next-page-tester`.

### Extra note

`@testing-library/dom` doesn't provide specific API for hydration but a `hydrate` option to provide to its usual render method.

### Naming things

I've currently introduced a few new API's, this is the current state:

```
const { renderHTML, render, html, page } = await getPage({ route });
const nextRoot = renderHTML();
const nextRoot = render();

// or:
const { nextRoot } = renderHTML();
const { nextRoot } = render();
```

// or:
const { container } = renderHTML();
const { container } = render();
```
This might be the right time to do one of the 3 most difficult things about coding: naming stuff :) Do you see any improvement we might do in this sense?

**Comment by Meemaw** [comment]
Regarding the `screen` and `within`: I've seen a pattern with utility test flows are quire popular out there. For example consider the following function:

```js
const testLoginFlow = () => {
   userEvent.type(screen.getByPlaceholderText('email'), 'user@example.com');
   userEvent.type(screen.getByPlaceholderText('password'), 'superHardPassword');

   userEvent.click(screen.getByText('login'));
   
   await screen.findByText('Logged in');
}

This function is then used in tons of tests and it doesn't have any dependencies (its using "global" `screen`). In ideal case, users would still be able to do that without having to pass `screen` bound from `within` all over the place.

**Comment by toomuchdesign** [comment]
This makes very sense. We should try and dig a bit into testing-library to see how we can integrate with global screen.

**Comment by toomuchdesign** [comment]
Global `screen` seems to be evaluated on module load:
https://github.com/testing-library/dom-testing-library/blob/f627ade1d8169db6113181a3893c847c10fa3ee2/src/screen.js#L41

Since our `renderHTML` method replaces `html` element content I expect `screen` to keep the reference to the old `body` element. Let's investigate.

We might consider updating `body` and `head` contents instead of replacing `html` content, but It doesn't sound like a great solution.

**Comment by toomuchdesign** [comment]
Now `renderHTML` preserves `document.body` identity and global TL `screen` object should work (I added a specific test for it):
https://github.com/toomuchdesign/next-page-tester/pull/83/commits/7f514d31e276d0a44f4a78f6053166df4690d3cc

File: src/getPage.tsx [hunk 1]
```diff
@@ -143,5 +149,6 @@ export default async function getPage({
     // @NOTE Temporary hack to keep tests green
     page: useDocument ? serverPageElement : clientPageElement,
```
Meemaw, line 150: Is this "hack" still needed?

src/getPage.tsx
```diff
@@ -40,7 +41,12 @@ export default async function getPage({
   router = (router) => router,
   useApp = true,
   useDocument = false,
-}: Options): Promise<{ page: React.ReactElement; html: string }> {
+}: Options): Promise<
+  {
```
Meemaw, line 45: Let's extract this into a type (e.g. `GetPageResult`)  and export it from library as well? It might be useful for people trying to wrap "getPage" with some default options.

src/makeRenderMethods.ts
```diff
@@ -0,0 +1,48 @@
...
+  html: string;
+  pageElement: JSX.Element;
+}): {
+  renderHTML: () => { nextRoot: HTMLElement };
+  render: () => { nextRoot: HTMLElement };
+} {
+  // Update whole document content with SSR html
+  // @NOTE we have to preserve document.body element identity
+  // to not break @testing-library global "screen" object
+  function renderHTML() {
```
Meemaw, line 17: I guess it's a bit confusing how this differs from `render`. Both render HTML at the end of the day Could we come up with a better name? For some reason Im thinking about `prerender`.

toomuchdesign: The difference consists of `renderHTML` rendering just plain HTML in the DOM while `render` also binds React and eventListeners to it. Let's take a moment to find out better names 👍

src/getPage.tsx
toomuchdesign: I'm currently working on a 3rd branch to integrate document render into the new rendering flow.

src/getPage.tsx
toomuchdesign: You mean to avoid typing:

```js
import { getPage } from 'next-page-tester';
type GetPageResult = ReturnType<typeof getPage>
```
?

Meemaw: Yeah. I know this is possible, but most libraries import this for ease of usage.

e.g.

```js
import { RenderResult } from '@testing-library/react';
```

Meemaw (approved):

PR #90, opened 2021-01-05
feat: render html server side and hydrate in JSDOM

toomuchdesign:

Refactor

`next-page-tester` has an important architectural limitation: it currently never performs a full tree render in server mode (like Next.js obviously does during SSR).

This process is skipped in favour of directly (and roughly) return an element tree similar to the one generated in-browser after React hydration. This solution works fine as long as we consider `_app` and pages.

## What is the new behaviour?

`getPage` does the following

- Fetch route data as usual
- Create 2 component trees of the whole page (root element: `#__next` div): one emulating the way it should be generated server side, and one client side

..and returns:
- `render` method which calls `renderHTML ` and hydrate `App` component to `__next` root element  _(it emulates Next.js app hydration when page is loaded)_
- `serverRender` method which replaces current DOM with the SSRed HTML _(it emulates Next.js SSR result)_
- `serverRenderAsString` method which returns the html string representing the SSR-rendered page element
- `page` element, client page element meant to be mounted (with React.hydrate) on `#__next` div

Yes. User should call `getPage` as usual and then render to the DOM using the returned **render methods**: `render` or `serverRender`.

Manually render `page` element (eg. with @testing-library/react) is currently still possible even if not suggested (since it has limitations and can produce results which are less trustable then rendering with the aforementioned render methods).

`page` element will always return Next.js root element (`#__next` div ) even when `useDocument` is enabled
## TODO

- [x] Reshape/rename `new-returns` tests
- [x] Update docs
- [x] Find alternative names for `renderHTML`, `render`, `html` and `page` return
- [x] ~Consider removing [document rendering setup](https://github.com/toomuchdesign/next-page-tester/blob/c4cc170414d6cd9c3fc6daa3c0502b65fbee83cb/src/_document/render.tsx#L30)~
- [x] Update root `<html/>` attributes in `rederHTML` (and uncomment related tests)
- [x] Add a `cleanup` method to run `afterEach` test (let's see how `@testing-library` hooks into global `afterEach`)
- [x] ~Consider removing `page` and/or `html` return value (is their value worth the price of extending user API?)~


src/_document/render.tsx
```diff
@@ -19,16 +19,27 @@ export default async function renderDocument({
...
+  }
+
+  const customDocumentFile = getDocumentFile({ options });
+  const Document = customDocumentFile.server.default;
 
   let head: JSX.Element[] = [];
 
   const renderPage: RenderPage = () => {
-    const html = renderToStaticMarkup(
+    const html = renderToString(
```
Meemaw, line 42: Have you tried `reusing` the output of this instead of rendering again in `getPage`? I think this function could simply return html `string`.

If `useDocument=false` we just `ReactDOM.renderToString` above.

toomuchdesign: Didn't intend to ignore your hint, I'm just still a bit confused about how document rendering works. :) I've opened PR #91 to streamline this part.

toomuchdesign: The `html` generated here seems only include the page's tree from `App` component down. We'd still need to generate the `html` for upper tree. I might definitely be missing something though.

Meemaw (approved):

README.md
```diff
@@ -6,19 +6,19 @@
 
 The missing DOM integration testing tool for [Next.js][next-github].
 
-Given a Next.js route, this library will return an instance of the matching page component instantiated with the **properties** derived by Next.js' [**routing system**][next-docs-routing] and [**data fetching methods**][next-docs-data-fetching].
+Given a Next.js route, this library will **render the matching page in JSDOM**, provided with the expected **props** derived from Next.js' [**routing system**][next-docs-routing] and [**data fetching methods**][next-docs-data-fetching].
 
 ```js
-import { render, screen, fireEvent } from '@testing-library/react';
 import { getPage } from 'next-page-tester';
+import { screen, fireEvent } from '@testing-library/dom';
```
Meemaw, line 13:
```suggestion
import { screen, fireEvent } from '@testing-library/react';
```

README.md
```diff
@@ -99,15 +125,19 @@ Until **Jest v27** is published, you might need to patch `jest` in order to load
...
 
+## Examples
+
+Under [examples folder][examples-folder] we're documenting the testing cases which `next-page-tester` enables.
+
 ## Notes
 
 - Data fetching methods' context `req` and `res` objects are mocked with [node-mocks-http][node-mocks-http]
-- Next page tester can be used with any testing framework/library (not tied to Testing library)
+- Next page tester can be used with any testing framework/library (it's not tied to Jest or Testing library)
```
Meemaw, line 135: Should we mention that library hasn't been tested with other test runners than `jest` and issues with those should be reported.

README.md
toomuchdesign: Nice one!

README.md
toomuchdesign: ✅

toomuchdesign:
I just found out that importing `@testing-library/dom` instead of `'@testing-library/react` somehow causes `@testing-library/user-event` to not wrap events into React `act()` wrapper. Therefore a lot of:

```
Warning: An update to RouterProvider inside a test was not wrapped in act(...)
```

👀

Meemaw:
I use `@testing-library/react` with `@testing-library/user-event` in all projects and it works without any problems (never importing `@testing-library/dom`) . I would recommend using `click` from `@testing-library/user-event` in our examples as well as its just more closely resembles "user" click.

toomuchdesign:
Yep. I though we could use `@testing-library/dom` for all those cases where we didn't need to perform a custom render. But it's probably not the case. 👍

Well. It's been a quite long journey, but I'm very positively surprised by the result we've been able to get to. I'm considering merging to master. 🚂

Meemaw:
I think its ready 🎉 Do you plan to release it as `1.0.0`?

The only concern I have is regarding the "jest" patch. Did all those changes make it "mandatory" for most use cases or is it still only needed for special "document" cases.

toomuchdesign:
I was actually thinking to stay in v0 for at least this first release. NPM and Yarn handle v0 in semver-expected way.

As for the Jest patch, I expect basic use cases to work anyway since imported modules identity should not be broken. Let's see what tests say.

Meemaw:
I would expect all these on the list to fail so I think we're good.

PR #91, opened 2021-01-05
[next] refactor: document render cleanup

toomuchdesign:

Refactor

Default document component is declared as a custom component in `src/_document/DefaultDocument.tsx`

## What is the new behaviour?

Use original Next.js default document exposed in `next/document`

No.


src/_document/getDocumentFile.ts
```diff
@@ -26,6 +25,6 @@ export default function getDocumentFile({
 
 function getDefaultDocumentFile(): PageFile<NextDocumentFile> {
   return loadFile<NextDocumentFile>({
-    absolutePath: path.resolve(__dirname, 'DefaultDocument.tsx'),
+    absolutePath: 'next/document',
```
Meemaw, line 28: Do we intentionally import it like this to resolve it in "server" environment?

Meemaw (approved):

src/_document/getDocumentFile.ts
toomuchdesign: Oui, maybe It's overkill?

Meemaw: Could be, but perhaps it's a safer option.

PR #92, opened 2021-01-05
[next] Update docs

toomuchdesign:

Docs update

If you find the time you're welcome to add anything you think we should mention @Meemaw.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ describe('Blog page', () => {
 
 ## What
 
-The idea behind this library is to reproduce as closer as possible the way Next.js works without spinning up servers, and render the output in a local JSDOM environment.
+The idea behind this library is to reproduce as closely as possible the way Next.js works without spinning up servers, and render the output in a local JSDOM environment.
 
 In order to provide a valuable testing experience `next-page-tester` replicates the **rendering flow of an actual `next.js` application**:
 
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ const { render, renderHTML, html, page } = await getPage({ options });
 Type: `() => HTMLElement<NextRoot>`<br/>
 Returns: `#__next` root element element.
 
-You will mostly use this render method. It calls `renderHTML()` and then mounts/hydrates the React application into JSDOM `#__next` root element. This is what the user gets when she visits a page.
+Unless you have an advances use-case, you should mostly use this method. Under the hood it calls a lower level `renderHTML()` method and then mounts/hydrates the React application into JSDOM `#__next` root element. This is what the user would get/see when they visit a page.
 
 ### renderHTML()
 
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ const { render, renderHTML, html, page } = await getPage({ options });
 Type: `() => HTMLElement<NextRoot>`<br/>
 Returns: `#__next` root element element.
 
-Unless you have an advances use-case, you should mostly use this method. Under the hood it calls a lower level `renderHTML()` method and then mounts/hydrates the React application into JSDOM `#__next` root element. This is what the user would get/see when they visit a page.
+Unless you have an advanced use-case, you should mostly use this method. Under the hood it calls `renderHTML()` and then mounts/hydrates the React application into JSDOM `#__next` root element. This is what users would get/see when they visit a page.
 
 ### renderHTML()
 
--- a/README.md
+++ b/README.md
@@ -53,32 +53,42 @@ The mounted application is **interactive** and can be tested with any DOM testin
 
 ### getPage
 
-`getPage` accepts an [option object](/#options) and returns 4 values:
+`getPage` accepts an [option object](#options) and returns 4 values:
 
-```ts
-function getPage(options: Options): Promise<
-  renderHTML: () => HTMLElement<NextRoot>
-  render: () => HTMLElement<NextRoot>
-  html: string
-  page: React.ReactElement<AppElement>
->;
+```js
+import { getPage } from 'next-page-tester';
+const { render, renderHTML, html, page } = await getPage({ options });
 ```
 
-#### `renderHTML()`
+### render()
 
-Inject server side rendering output to the DOM. Returns `#__next` root element element.
+Type: `() => HTMLElement<NextRoot>`<br/>
+Returns: `#__next` root element element.
 
-#### `render()`
+You will mostly use this render method. It calls `renderHTML()` and then mounts/hydrates the React application into JSDOM `#__next` root element. This is what the user gets when she visits a page.
 
-Calls `renderHTML()` and mounts/hydrate the the application on `#__next` root element. Returns `#__next` root element element.
+### renderHTML()
 
-#### `html`
+Type: `() => HTMLElement<NextRoot>`<br/>
+Returns: `#__next` root element element.
+
+Inject the output of server side rendering into the DOM and doesn't mount React. Use it to test how Next.js renders in the following scenarios:
+
+- before Reacts mounts
+- when JS is disabled
+- SEO tests
+
+### html
+
+Type: `string`
 
 HTML string representing output of server side rendering.
 
-#### `page`
+### page
 
-React element of the application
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,19 +6,19 @@
 
 The missing DOM integration testing tool for [Next.js][next-github].
 
-Given a Next.js route, this library will return an instance of the matching page component instantiated with the **properties** derived by Next.js' [**routing system**][next-docs-routing] and [**data fetching methods**][next-docs-data-fetching].
+Given a Next.js route, this library will **mount the matching page in JSDOM**, provided with the expected **props** derived from Next.js' [**routing system**][next-docs-routing] and [**data fetching methods**][next-docs-data-fetching].
 
 ```js
-import { render, screen, fireEvent } from '@testing-library/react';
```
Meemaw, line 12: Would it make sense to make this a section with "Using @testing-library/react" header and provide another example how library could be used without it?

README.md
```diff
@@ -29,20 +29,57 @@ describe('Blog page', () => {
...
+>;
+```
+
+#### `renderHTML()`
+
+Inject server side rendering output to the DOM. Returns `#__next` root element element.
+
+#### `render()`
+
+Calls `renderHTML()` and mounts/hydrate the the application on `#__next` root element. Returns `#__next` root element element.
```
Meemaw, line 73: Lets clarify when user would want to use each option. If I understand it correctly, `render` should be used in most cases, but `renderHTML` could be used if users want more control and do the `hydration` themselves?

README.md
```diff
@@ -29,20 +29,57 @@ describe('Blog page', () => {
...
+
+#### `html`
+
+HTML string representing server side rendering output.
+
+#### `page`
+
+React element the application
```
Meemaw, line 81:
```suggestion
React element of the application
```

README.md
```diff
@@ -29,20 +29,57 @@ describe('Blog page', () => {
...
+
+Inject server side rendering output to the DOM. Returns `#__next` root element element.
+
+#### `render()`
+
+Calls `renderHTML()` and mounts/hydrate the the application on `#__next` root element. Returns `#__next` root element element.
+
+#### `html`
+
+HTML string representing server side rendering output.
```
Meemaw, line 77:
```suggestion
HTML string representing output of server side rendering.
```

README.md
```diff
@@ -29,20 +29,57 @@ describe('Blog page', () => {
 
 ## What
 
-The idea behind this library is to enable DOM integration tests on Next.js pages along with [data fetching][next-docs-data-fetching] and [routing][next-docs-routing].
+The idea behind this library is to reproduce as closer as possible the way Next.js works without spinning up servers, and render the output in a local JSDOM environment.
```
Meemaw, line 32:
```suggestion
The idea behind this library is to reproduce as closely as possible the way Next.js works without spinning up servers, and render the output in a local JSDOM environment.
```

README.md
toomuchdesign: Yep `hydration` themselves but especially test what a search engine would see. 👍

README.md
toomuchdesign: What if we did the opposite? :) Like adding a section where we explain that `next-page-tester` does nothing more then appending stuff in in the (mocked) DOM and call React.hydrate on it.

I assume `@testing-library` is almost a de-facto standard library and I'd like to not clutter the initial example with too 
many DOM API's (considering we simulate a click event). How do you see it?

Meemaw: Perhaps we should create an `examples` folder where we can "gather" some patterns to not clutter `README.md`. It can be an ongoing effort as a "response" to potential issues users might face.

README.md
```diff
@@ -6,19 +6,19 @@
 
 The missing DOM integration testing tool for [Next.js][next-github].
 
-Given a Next.js route, this library will return an instance of the matching page component instantiated with the **properties** derived by Next.js' [**routing system**][next-docs-routing] and [**data fetching methods**][next-docs-data-fetching].
+Given a Next.js route, this library will **mount the matching page in JSDOM**, provided with the expected **props** derived from Next.js' [**routing system**][next-docs-routing] and [**data fetching methods**][next-docs-data-fetching].
```
Meemaw, line 9:
```suggestion
Given a Next.js route, this library will **render the matching page in JSDOM**, provided with the expected **props** derived from Next.js' [**routing system**][next-docs-routing] and [**data fetching methods**][next-docs-data-fetching].
```

README.md
```diff
@@ -29,20 +29,67 @@ describe('Blog page', () => {
...
+import { getPage } from 'next-page-tester';
+const { render, renderHTML, html, page } = await getPage({ options });
+```
+
+### render()
+
+Type: `() => HTMLElement<NextRoot>`<br/>
+Returns: `#__next` root element element.
+
+You will mostly use this render method. It calls `renderHTML()` and then mounts/hydrates the React application into JSDOM `#__next` root element. This is what the user gets when she visits a page.
```
Meemaw, line 68:
```suggestion
Unless you have an advances use-case, you should mostly use this method. Under the hood it calls a lower level `renderHTML()` method and then mounts/hydrates the React application into JSDOM `#__next` root element. This is what the user would get/see when they visit a page.
```

README.md
```diff
@@ -29,20 +29,67 @@ describe('Blog page', () => {
...
+
+### renderHTML()
```
Meemaw, line 70: What do you think about this?

```suggestion
### prepareDocument()
```

toomuchdesign: I'll try to put down what I think `renderHTML` does compared to a real Next.js app.

From my point of `renderHTML()/whatever` represents the usual initial rendering process which happens in any browser when a user visits any HTML page: HTML file download, parsed and a DOM tree generated out if it.

For a static HTML page the game ends with `renderHTML()/whatever`.

The exact same thing happens for a SSRendered React app but there's is an extra step: when JS resources are loaded React takes over, and does whatever it was told to do with the just rendered DOM.

From users' perspective, I'd say that `renderHTML()` renders the `html` string in the DOM (or any equivalent word for transforming a string into it's visual representation)

`prepareDocument` sounds a bit like "setting up the environment before we do anything", while `renderHTML()` paints the whole page markup on the screen (except for the detail that JSDOM doesn't provide visual output 😄 ).

`renderHTML()` should be something close to:
- `renderSSRResultInDOM`
- `renderSSR`
- `renderStaticHTML`
- `renderSSRHTML`

`render()` should be something like:
- `renderAndMountPage`
- `renderPageAndMountReact`
- `renderSSRAndMountPage`
- `renderAndHydrate`

toomuchdesign:
Extra consideration:

I suspect that we might want to rename also `html` and `page` returned from `getPage`. Especially `page` sounds quite a confusing word in current context.

If I'd optimistically expect `html` to be a string, it's not clear what `page` means. Currently it is something very specific: that is the React element (executed and rendered in browser environment) representing the `_app` component meant to be hydrated  into SSRed `__next` element.

README.md
Meemaw: `renderStaticMarkup` & `render`?

I think `renderStaticMarkup` makes it clear that output is SSR and that it's not interactive. As for the "main" `render` method I would not complicate and make name as simple as possible: `render`.

toomuchdesign: I'd slim down the list to:
- renderStaticMarkup
- renderStaticDocument
- renderDocument

Taken from React docs:
> Render a React element to its initial HTML. React will return an HTML string. You can use this method to generate HTML on the server and send the markup down on the initial request for faster page loads and to allow search engines to crawl your pages for SEO purposes.

> If you call ReactDOM.hydrate() on a node that already has this server-rendered markup, React will preserve it and only attach event handlers, allowing you to have a very performant first-load experience.

- renderInitialHTML
- renderInitialMarkup
- renderServerRenderedMarkup
- renderServerRenderedHTML

(even though our server-rendered is emulated)

I'd try not to add another return method, right now. I don't see much value right now. Actually I considered removing both `html` and `page`: they have sense only if someone tried to render and hydrate on their side.

PR #93, opened 2021-01-05
[next] Render root html element attributes

toomuchdesign:

Feature

JSDOM root `html` element is not updated with `html` attributes returned from SSR render (this should happen in `renderHTML`)

## What is the new behaviour?

Parse the `html` string with an XML parsee, retrieve `html` elements attribute and manually append them to JSDOM `html` element.

## Other information:

Using an XML parser to get these few attributes seems overkill, but I haven't been able to find a better way to update JSDOM `html` root element attributes.

@Meemaw if a simpler solution came to your mind, I'm all ears. I feel like there has to be a better way, but...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,6 +46,18 @@
           "integrity": "sha1-tdRU3CGZriJWmfNGfloH87lVuv0=",
           "dev": true
         },
+        "htmlparser2": {
+          "version": "5.0.1",
+          "resolved": "https://registry.npmjs.org/htmlparser2/-/htmlparser2-5.0.1.tgz",
+          "integrity": "sha512-vKZZra6CSe9qsJzh0BjBGXo8dvzNsq/oGvsjfRdOrrryfeD9UOBEEQdeoqCRmKZchF5h2zOBMQ6YuQ0uRUmdbQ==",
+          "dev": true,
+          "requires": {
+            "domelementtype": "^2.0.1",
+            "domhandler": "^3.3.0",
+            "domutils": "^2.4.2",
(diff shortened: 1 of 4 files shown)


Meemaw:
I've quickly looked into it and it seems attributes are lost in `renderHtml` when `document.documentElement.innerHTML` is set. If you `console.log` html string at that point you will see it has all the attributes.

As I see it issue is that `document.documentElement` is the "root" html and not the one we want to "set".

toomuchdesign:
Yep, the issue is in `renderHTML` where we cannot set `outerHTML` on root `html` element.

If we can't find a cheaper way of extracting `html` attributes from the string, the choice consists of which parser to use: `JSDOM DOMParser` (which uses `parse5` under the hood) or another one (I took `fast-xml-parser`).

### JSDOM DOMParser
Pro: no extra dependency, only DOM api
Cons: it parses a string into a full DOM tree (which we don't actually need)

### fast-xml-parser
Pro: it parses a string into a simple object. It's configurable so that it stop parsing when it encounters `body` and `html elements
Cons: 1 extra dependency, specific parser API

My only (little) concern here is parsing `html` string twice: on `innerHTML` set first and to retrieve `html` element attributes.

`fast-xml-parser` would reduce the second parsing operations to a bare minimum (especially for large documents since it doesn't parse head and body), but I don't have evidence of any performance gain, yet.

toomuchdesign:
Found a solution involving DOM API's trickeries only and just one parsing pass. 👍 Thanks for the tip @Meemaw!

PR #94, opened 2021-01-06
test(third-party): enzyme

Meemaw:

Test/example for `enzyme` that serves as a quick check that it is indeed possible to use `enzyme` with `next-page-tester`.

This is a very basic example as "enzyme" should not be promoted by any means, but unfortunately a lot of users still use it.


toomuchdesign:
What if we transformed this into an example in plain markdown (we might add an example folder)? A part the extra dev dependencies, I think that documenting `enzyme` usage **as test** doesn't feel fully right since we don't actively support it (expect for returning a React element), and if in future enzyme tests turned red, we wouldn't probably take any action.

toomuchdesign (commented):
What if we transformed this into an example in plain markdown (we might add an example folder)? A part the extra dev dependencies, I think that documenting `enzyme` usage **as test** doesn't feel fully right since we don't actively support it (expect for returning a React element), and if in future enzyme tests turned red, we wouldn't probably take any action.

Meemaw:
This test should be passing as long as we are exporting `page` from the `getPage`. We are testing that page element works correctly through other means so it shouldn't be a problem. With this I just wanted to confirm that it indeed works. Fell free to transform it into another shape if you think it makes more sense. I'm just afraid of having examples that we are not testing.

toomuchdesign:
hat id we added add a test case for returned `page` in `get-page-return` tests to ensure `page` is a React element? That's all we guarantee to Enzyme, after all, right?

Actually only `page` is missing in `get-page-return` tests.

toomuchdesign:
Closed in favour of #97.

PR #95, opened 2021-01-06
test(third-party): styletron-react

Meemaw:

Third party test for `styletron-react`.

Replaces https://github.com/toomuchdesign/next-page-tester/pull/76/ and now works without any hacks which is a great step forward 🚀 

Edit: nevermind, we still have 1 missing part in our `_document` implementation that is causing the `warnings` here.


src/__tests__/third-party/styletron-react/pages/_document.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { styletron } from '../styletron';
+
+type Props = RenderPageResult & {
+  stylesheets: Sheet[];
+};
+
+export default class CustomDocument extends Document<Props> {
+  static async getInitialProps(ctx: DocumentContext) {
+    const page = ctx.renderPage((App) => (props) => (
+      <StyletronProvider value={styletron}>
```
Meemaw, line 21: This code is never executed. We are missing this part from `NextJS` https://github.com/vercel/next.js/blob/canary/packages/next/next-server/server/render.tsx#L961 which doesn't seem too complicated, but is hard to add into our code.

src/__tests__/third-party/styletron-react/styletron-react.test.ts
```diff
@@ -0,0 +1,22 @@
+import { getPage } from '../../../index';
+import userEvent from '@testing-library/user-event';
+import { screen } from '@testing-library/react';
+
+describe('styletron-react', () => {
+  it('As a user I can test applications using "styletron-react"', async () => {
+    // There are some warnings/errors due to server & client className missmatch
+    // They don't seem to be a bug in "next-page-tester" but un-complete implementation of stylesheets in JSDOM
```
Meemaw, line 8: They are bug in "next-page-tester" 🙂

src/__tests__/third-party/styletron-react/pages/_document.tsx
toomuchdesign: Are these errors triggered because of this `next-page-tester` missing implementation or the normally happen in a real app?

toomuchdesign:
I'm sure I'm missing context here, but I'll make the question anyway :) What styletron provides is a mean to style components, right?

Since we do DOM testing, wouldn't a user mock `styletron` out? Which is what already happens with imported CSS files. 

The best would be being able to render any Next.js setup seamlessly (I agree and maybe we'll be able to get there), but is it much of a problem not to be able to JSDOM render `css-in-js` libraries setups? I'm asking since I'm probably missing the whole picture.

Meemaw:
`styletron` is just one example of  many `css-in-js` libraries. Perhaps it is more popular in server rendered react applications due to atomic classNames where application bundle size is crucial.

I use https://github.com/uber/baseweb in my NextJS projects which uses `styletron-react` under the hood, thus the motivation for adding this specific test. As you can see from the custom `_document`, setup is not so trivial and thus a good example for a "third-party" test.

With the current setup there are warning and errors being thrown away as we miss implementation for [customizing renderPage](https://nextjs.org/docs/advanced-features/custom-document#customizing-renderpage). It is not critical at all, since HTML markup that users care about and test is correct in this case, but it can be annoying.

I imagine those warnings/errors would pop up with other `css-in-js` libraries as well (`styled-components`,...) which are way more popular.

I think we can ignore this for now and prioritise getting "next" branch on master. I can tackle this after.

PR #96, opened 2021-01-07
[next] Rename return object

toomuchdesign:

Refactor

## What is the new behaviour?

Rename `getPage` return values.

Current API:

```js
const { render, renderHTML, html, page } = await getPage({ options });
```

New API:

```js
const { render, serverRender, serverRenderToString, page } = await getPage({ options });
```

Yes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,14 +62,14 @@ const { render, serverRender, html, page } = await getPage({ options });
 
 ### render()
 
-Type: `() => HTMLElement<NextRoot>`<br/>
+Type: `() => { nextRoot: HTMLElement<NextRoot> }`<br/>
 Returns: `#__next` root element element.
 
 Unless you have an advanced use-case, you should mostly use this method. Under the hood it calls `serverRender()` and then mounts/hydrates the React application into JSDOM `#__next` root element. This is what users would get/see when they visit a page.
 
 ### serverRender()
 
-Type: `() => HTMLElement<NextRoot>`<br/>
+Type: `() => { nextRoot: HTMLElement<NextRoot> }`<br/>
 Returns: `#__next` root element element.
 
 Inject the output of server side rendering into the DOM and doesn't mount React. Use it to test how Next.js renders in the following scenarios:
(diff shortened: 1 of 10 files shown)


toomuchdesign:
Moved naming topic in a separate PR.

I gave `serverRender` a try, since it's exactly what we try to do here.

I though that "static" could not be totally appropriate, after all, since If React decided to return some SSR markup which includes (eg. inline) event binding, it wouldn't be static anymore, right? Therefore it doesn't really have to be static markup. 🤯 

Maybe it's the right opportunity to also rename `html` and `page`.

`html` could become:

- initialHTML
- serverHTML
- initialMarkup
- serverMarkup
- stay `html` :)

Meemaw:
> Moved naming topic in a separate PR.
> 
> I gave `serverRender` a try, since it's exactly what we try to do here.

I think this is okay for now. Perhaps we are just over-thinking but this library is still in early phases and this can be changed sometime in the future with a breaking release if absolutely needed.

> Maybe it's the right opportunity to also rename `html` and `page`.
> 
> `html` could become:
> 
> * initialHTML
> * serverHTML
> * initialMarkup
> * serverMarkup
> * stay `html` :)

I'm fine with any. I think `html` is okay for now.

Extra: I think we can avoid exporting `html` from `getPage` and return it from `serverRender`. This would also better streamline responsibilities of different functions.

Name `getPage` doesn't indicate any rendering. It just prepares the primitives to work with. `html` is available only after user "renders" the page. Internally this means we can move this code:

```js
 const html: string = executeAsIfOnServerSync(() =>
    ReactDOMServer.renderToString(documentElement)
  );
```

Into `serverRender`

This means new API would look like this:
```js
const { render, serverRender, page } = await getPage({ options });
const { html } = serverRender();
```

Meemaw (approved):

toomuchdesign:
```js
const { html } = serverRender();
```

...is an elegant API. I seem to see a small downside, though. Since `serverRender` operates over the DOM to apply `html`, users would loose the opportunity to get `html` and `page` to skip our render methods and interact with the DOM on their own with their custom logic. Isn't it?

Meemaw:
Hmm, yeah. Perhaps we should mention in `README.md` that `serverRender` has side-effects and will modify global document objects.

toomuchdesign:
Ok, let's move `renderToString` out of `getPage`. In order to avoid side effects we might return a 3rd render method which renders to string. Or live with the side/effect and return it from `serverRender`.

toomuchdesign:
`getPage` now returns and extra render method: `renderServerToString`. We expose and extra piece of API but now the responsibilities of each render method are quite nicely defined. How do you like it @Meemaw?

Meemaw:
Looks good @toomuchdesign 🚀

PR #101, opened 2021-01-11
feat: catch errors & render _error page

Meemaw:

Feature:
- Implement ErrorPage rendering
- Add support for custom `_error` page

Error thrown in user code are not catched and propagated. This doesn't reflect what the output would be in a real situtation for a user and makes it impossible to test "bad scenarios".

## What is the new behaviour?

Error page is rendered

Mostly likely no. Unless someone is catching exceptions in their tests.

### DIFF
--- a/src/__tests__/page-extensions/page-extensions.test.tsx
+++ b/src/__tests__/page-extensions/page-extensions.test.tsx
@@ -16,15 +16,16 @@ describe('page file extensions', () => {
     });
 
     describe('unknown extension', () => {
-      it('throws "page not found" error', async () => {
-        await expect(
-          getPage({
-            nextRoot,
-            route: '/invalid',
-          })
-        ).rejects.toThrow(
-          '[next-page-tester] No matching page found for given route'
-        );
+      it('throws "page not found" error page', async () => {
+        const { render } = await getPage({
+          nextRoot,
+          route: '/invalid',
+        });
+        render();
+        expect(screen.getByText('404')).toBeInTheDocument();
+        expect(
+          screen.getByText('This page could not be found', { exact: false })
+        ).toBeInTheDocument();
       });
     });
   });
@@ -42,15 +43,17 @@ describe('page file extensions', () => {
       });
     });
     describe('not allowed extensions', () => {
-      it('throws "page not found" error', async () => {
-        await expect(
-          getPage({
-            nextRoot,
-            route: '/js',
-          })
-        ).rejects.toThrow(
-          '[next-page-tester] No matching page found for given route'
-        );
+      it('throws "page not found" error page', async () => {
+        const { render } = await getPage({
+          nextRoot,
+          route: '/js',
+        });
+        render();
+
+        expect(screen.getByText('404')).toBeInTheDocument();
+        expect(
+          screen.getByText('This page could not be found', { exact: false })
+        ).toBeInTheDocument();
       });
     });
   });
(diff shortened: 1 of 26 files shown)


src/_document/render.tsx
```diff
@@ -8,6 +8,17 @@ import { renderToString } from 'react-dom/server';
 import { HeadManagerContext } from 'next/dist/next-server/lib/head-manager-context';
 import type { DocumentProps } from 'next/document';
 
+export const wrapWithDocument = (pageElement: JSX.Element) => {
```
toomuchdesign, line 11: wrapWithEmptyDocument?

src/__tests__/use-app/use-app.test.tsx
```diff
@@ -149,15 +150,16 @@ describe('_app support', () => {
...
+    it('throws "page not found" error page', async () => {
+      const { render } = await getPage({
+        nextRoot: __dirname + '/__fixtures__/custom-app-with-gip',
+        route: '/_app',
+      });
+      render();
+      expect(screen.getByText('404')).toBeInTheDocument();
+      expect(
+        screen.getByText('This page could not be found', { exact: false })
+      ).toBeInTheDocument();
```
toomuchdesign, line 162: If we went this direction, we might want to wrap up a test utility function to say "expect this to be error page".

src/__tests__/page-extensions/page-extensions.test.tsx
```diff
@@ -16,15 +16,16 @@ describe('page file extensions', () => {
...
+      it('throws "page not found" error page', async () => {
+        const { render } = await getPage({
+          nextRoot,
+          route: '/invalid',
+        });
+        render();
+        expect(screen.getByText('404')).toBeInTheDocument();
+        expect(
+          screen.getByText('This page could not be found', { exact: false })
+        ).toBeInTheDocument();
```
toomuchdesign, line 28: Same as before: we might want to wrap up a test utility function to say "expect this to be not found page".

Meemaw:
This is just a proof of concept I initially started in https://github.com/toomuchdesign/next-page-tester/pull/88.

It is by no means complete yet:
- doesn't work after client side navigation
- This PR could also introduce support for `404` page as code seems related to this
- Refactors/cleanup

I will be quite busy this week so fell free to work on it if you find some time @toomuchdesign

toomuchdesign:
Just dropping a note: if on one hand we should aim to replicate what Next.js does (and I see a lot of cases where rendering the expected error or 404 page is clearly an advantage), on the other i wonder whether a user might prefer receiving an error about a route not matching any page. How do you see it?

toomuchdesign:
No worries @Meemaw, I'm also going to be considerably busier from now. We definitely did more then I expected, here 🚀 .
I'll check your PR out and try to focus on implementing next/config support.

Meemaw:
> Just dropping a note: if on one hand we should aim to replicate what Next.js does (and I see a lot of cases where rendering the expected error or 404 page is clearly an advantage), on the other i wonder whether a user might prefer receiving an error about a route not matching any page. How do you see it?

I think we had similar discussion for the `redirect` where an option was to return `redirect` object instead of do it behind the scenes. It just depends where you position the library, e.g. unit or integration testing.

If "error" is returned we are closer to unit testing and would be testing implementation details as opposed to what the "end users" care about. As an end user of an app I don't care if Error is thrown in the app, but how the error is handled, e.g. the error screen I will see.

I personally see this library at the integration level with a strong synergy with @testing-library/react where everything should is tested as it would appear on screen.

toomuchdesign:
Yep, I substantially agree.

I wanted to make sure we are aware of the design decisions we're making. Even thought I'm not a fan of giving flexibility by adding options, I'll try to see whether there's room to optionally throw when a route is not found. Or abandon the idea 👍

toomuchdesign:
From [Next.js docs](https://nextjs.org/docs/advanced-features/custom-error-page):

> pages/_error.js is only used in production. In development you’ll get an error with the call stack to know where the error originated from.

I put it here for reference.

PR #102, opened 2021-01-13
feature: support next/config runtime configuration

toomuchdesign:

Feature

`next/config` is not set and always returns `undefined`.

## What is the new behaviour?

Support [Next.js runtime configuration](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration) and set different `next/config` values based on `server` or `client` env. 

Yes.

### DIFF
--- a/src/setNextRuntimeConfig.ts
+++ b/src/setNextRuntimeConfig.ts
@@ -1,4 +1,4 @@
-import * as envConfig from 'next/dist/next-server/lib/runtime-config';
+import { setConfig } from 'next/dist/next-server/lib/runtime-config';
 import { getNextConfig } from './nextConfig';
 
 export default function setNextRuntimeConfig({
--- a/src/setNextRuntimeConfig.ts
+++ b/src/setNextRuntimeConfig.ts
@@ -9,7 +9,7 @@ export default function setNextRuntimeConfig({
   const config = getNextConfig();
   const { serverRuntimeConfig, publicRuntimeConfig } = config;
 
-  envConfig.setConfig({
+  setConfig({
     serverRuntimeConfig: runtimeEnv === 'server' ? serverRuntimeConfig : {},
     publicRuntimeConfig,
   });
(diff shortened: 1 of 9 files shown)


src/setNextRuntimeConfig.ts
```diff
@@ -0,0 +1,16 @@
+import * as envConfig from 'next/dist/next-server/lib/runtime-config';
```
Meemaw, line 1:
```suggestion
import { setConfig } from 'next/dist/next-server/lib/runtime-config';
```

src/setNextRuntimeConfig.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export default function setNextRuntimeConfig({
+  runtimeEnv,
+}: {
+  runtimeEnv: 'server' | 'client';
+}): void {
+  const config = getNextConfig();
+  const { serverRuntimeConfig, publicRuntimeConfig } = config;
+
+  envConfig.setConfig({
```
Meemaw, line 12:
```suggestion
  setConfig({
```

src/setNextRuntimeConfig.ts
Meemaw, line 1: Would it make more sense to keep this function along with the others in one place? (`src/nextConfig`).

Meemaw (approved):
Looks good. Left some minor comments.

src/setNextRuntimeConfig.ts
toomuchdesign: My take here (I could have misunderstood Next.js), is that `next.config.js` and Next.js runtime configuration (`next/config`) are 2 unrelated moving parts of Next.js.

`next/config` depends on `next.config.js` as much as many other Next.js parts do. On our case we also use `next.config.js` to get the allowed file extensions.

But maybe they are closer than I though and keeping them together might make sense. How do you see it?

Meemaw: Agree those are 2 different parts. LGTM

PR #109, opened 2021-01-14
chore: eslint setup

Meemaw:

Chore: setup `eslint` to have a consistent coding style and avoid unnecessary bugs.


toomuchdesign (changes requested):
Hi @Meemaw!
In my solo projects I removed eslint to stay with Prettier only and I definitely have no regrets :)

BTW let's give pedantic eslint another try! I might adjust the configuration in future if it bothers too much. Just a few point, thought:

- does `concurrently` bring an actual benefit to the project?
- may we update renovate configuration to update all eslint related stuff in a single go? (this the actual change request)

Not related, I saw that mac testing environments usually takes way longer to complete. Do you think we might consider dropping macos tests relying on the hope (😄 ) that a Node app that breaks on macos should also break on linux?

Meemaw:
> Hi @Meemaw!
> In my solo projects I removed eslint to stay with Prettier only and I definitely have no regrets :)
> 
> BTW let's give pedantic eslint another try! I might adjust the configuration in future if it bothers too much. Just a few point, thought:

I tried to do a fairly un-opinionated setup and I think the changes do make sense (`unused-imports`, `prefer-const`, `no-ts-ignore`,...). Some rules can be annoying, but it's easy to turn them off.

> 
> * does `concurrently` bring an actual benefit to the project?

It makes things much faster on my machine and in CI as well for some of the projects I work on.

> * may we update renovate configuration to update all eslint related stuff in a single go? (this the actual change request)
> 

Yep, will do.

> Not related, I saw that mac testing environments usually takes way longer to complete. Do you think we might consider dropping macos tests relying on the hope (😄 ) that a Node app that breaks on macos should also break on linux?

Well it takes around 2-3 minutes and thats nothing in my opinion (and its free 🙂 ). I'm used to CIs running for ~1 hour.

Meemaw:
@toomuchdesign if you feel comfortable with this going forward I would merge it as soon as possible to avoid merge conflicts that will probably start to appear.

toomuchdesign (approved):

Issue #111, opened 2021-01-15
Cannot find module 'cookie'


jansepke:
## 🐛 Bug Report

I think with version 0.15.0 a dependency to the `cookie` package was introduced in `makeContextObject.ts` but not added as dependency to `package.json`.

## To Reproduce

directly happens on executing `initTestHelpers()`

## Expected behavior

`cookie` added as dependency.

## Run `npx envinfo --preset jest`

```bash
  System:
    OS: Linux 5.4 Ubuntu 20.04.1 LTS (Focal Fossa)
    CPU: (4) x64 Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz
  npmPackages:
    jest: 26.6.3 => 26.6.3 
```
Next.js version: 10.0.5

Meemaw:
Yeah, forgot to add `cookie` to dependencies, was somehow thinking it is a dependency of `nextjs`. Should be fixed in https://github.com/toomuchdesign/next-page-tester/pull/112

PR #112, opened 2021-01-15
fix: add cookie dependency

Meemaw:

Fixes https://github.com/toomuchdesign/next-page-tester/issues/111

`cookie` dependency is missing

## What is the new behaviour?

...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2970,9 +2970,9 @@
       }
     },
     "@types/cookie": {
-      "version": "0.3.3",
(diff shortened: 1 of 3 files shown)


toomuchdesign (commented):
Do we also need `@types/cookie`? (in devDependencies)

Meemaw:
We actually already have it through this chain:
`@aws-amplify/core` > `universal-cookie` > `@types/cookie`

I will add it regardless for consistency.

toomuchdesign (approved):

PR #117, opened 2021-01-17
chore: refactor v1

Meemaw:

Refactor.

Implementing some of new features like [custom error page](https://nextjs.org/docs/advanced-features/custom-error-page) and [customizing render page](https://nextjs.org/docs/advanced-features/custom-document#customizing-renderpage) require some refactoring.

Currently the main problem I'm facing is passing around excessive objects (prop drilling)  e.g. `"options" ` and `"pageObject"` that are actually not needed by the functions (only some fields are needed). When handling errors, page might throw early in the process (e.g. in data fetching methods) meaning we wont have all the data needed to construct those complex data.

We should aim to keep function parameters and return types to minimum (only what the functions need to receive/return). This will make them more reusable in different contexts where not all the data is available to construct whole objects.

## What is the new behaviour?

Small refactor that tries to fix some of the problems mentioned above. There are still many places where we are passing much more than actually needed, but they can be fixed as we go.

No

### DIFF
--- a/src/RouterProvider.tsx
+++ b/src/RouterProvider.tsx
@@ -1,39 +1,45 @@
 import React, { useState, useCallback, useRef } from 'react';
+
 import { RouterContext } from 'next/dist/next-server/lib/router-context';
 import makeRouterMock, { PushHandler } from './makeRouterMock';
 import { useMountedState } from './utils';
-import type { ExtendedOptions, Page, PageObject } from './commonTypes';
+import {
+  ExtendedOptions,
+  Page,
+  RuntimeEnvironment,
+  RouteData,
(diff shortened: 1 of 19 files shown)


toomuchdesign:
Hi @Meemaw, where is option object causing troubles?
Until now option object worked fine and I think helped a lot in keeping function signature and types under control.

Extra point: have you considered beginning with 404 page implementation which seems to be a subset of Error page complexity?

Meemaw:
Looking at it again, `options` object is not problematic. I've taken some of the changes from this PR and included them in https://github.com/toomuchdesign/next-page-tester/pull/120.

PR #119, opened 2021-01-17
feat: loading environment variables

toomuchdesign:

Feature

## What is the new behaviour?

Next.js allow setting custom environments variables via [`next.config.js`](https://nextjs.org/docs/api-reference/next.config.js/environment-variables) and [dotfiles](https://nextjs.org/docs/basic-features/environment-variables#loading-environment-variables).

Here we implement both ways according to Next.js specs.

The way we make these env vars available differs from Next.js implementation: Next.js, involving a build step, replaces `process.env.xxx` occurrencies with the expected values, while here we directly set the expected values to global `process.env` object and clean it up between tests.

Well, we actually replace `process.env` object with an an extended copy of it.

### Which dotfile

For the sake of proving the PR works, we're currently temporarily loading `.env.local` dot file, but it's not what a user would expect.

I'd like to not make an option for it, unless absolutely forced. How would you see considering only `.env.test` dotfile?

### Env vars cleanup

Since I'd expect out users to use this lib against the same application, maybe the env vars cleanup between tests is quite unnecessary. Env vars get already reset between different test files. 

Yes.

## Other information:

`@next/env` doesn't currently forward the [`log` prop](https://github.com/vercel/next.js/blob/v10.0.6-canary.4/packages/next-env/index.ts#L116) rendering multiple logs when a dotfile is found. It seems to be a tiny Next.js bug and we might try to issue a pr.

https://github.com/vercel/next.js/issues/21296


Meemaw:
> 
> The way we make these env vars available differs from Next.js implementation: Next.js, involving a build step, replaces `process.env.xxx` occurrencies with the expected values, while here we directly set the expected values to global `process.env` object and clean it up between tests.
> 
> Well, we actually replace `process.env` object with an an extended copy of it.

Does that mean environment variables from env/set from code wont work?

> 
> ## Missing points
> ### Which dotfile
> For the sake of proving the PR works, we're currently temporarily loading `.env.local` dot file, but it's not what a user would expect.
> 
> I'd like to not make an option for it, unless absolutely forced. How would you see considering only `.env.test` dotfile?

I think `env.test` could be the default, but we might have some fallbacks in case it doesn't exist? I personally never use `env.test` in my projects.

> 
> ### Env vars cleanup
> Since I'd expect out users to use this lib against the same application, maybe the env vars cleanup between tests is quite unnecessary. Env vars get already reset between different test files.

Good point. I don't think there is any harm in doing it TBH.

toomuchdesign:
> Does that mean environment variables from env/set from code wont work?

I expect it to work in most of scenarios (like setting env vars from command), since we shallow copy `process.env` after `getPage` is called.

> I think env.test could be the default, but we might have some fallbacks in case it doesn't exist? I personally never use env.test in my projects.

Like `.env*.test`?

Meemaw:
> > I think env.test could be the default, but we might have some fallbacks in case it doesn't exist? I personally never use env.test in my projects.
> 
> Like `.env*.test`?

I meant something like this. Try to load `.env.test` and if doesn't exist fallback to `.env.production`. If that one doesn't exist fallback to `.env.local`.

So `.env.test` > `.env.production` > `.env.local`  (not sure about the order of last 2)


Also I would introduce a `console.log` message about the env file that has been loaded. `nextjs` currently logs such message:
`Loaded env from /bla/bla/bla/bla/bla/bla/bla/.env.development`

☝️ This might avoid confusion for the user if he expected different file to be loaded.

Meemaw:
> > Does that mean environment variables from env/set from code wont work?
> 
> I expect it to work in most of scenarios (like setting env vars from command), since we shallow copy `process.env` after `getPage` is called.
> 

You could test it by setting something before the `getPage` call.

```js
process.env.VARIABLE_FROM_CODE = 'test'
const { render } =  await getPage()
...
```

Meemaw (approved):

src/setEnvVars.ts
```diff
@@ -0,0 +1,60 @@
...
+  info: () => {},
+  error: () => {},
+};
+
+let dotFile: Record<string, string> | undefined = undefined;
+export function loadDotFile({ nextRoot }: { nextRoot: string }) {
+  // Temporarily hide global process.env to prevent "@next/env" from mutating it
+  const currentProcessEnv = process.env;
+  process.env = {};
+  const { combinedEnv } = loadEnvConfig(nextRoot, undefined, Logger);
```
Meemaw, line 14: Should we add an assertion to our test that info about loaded file is loaded?

toomuchdesign: The issue with how `@next/env` works is that, since we cannot control logger output, a user would get a log every time `getPage` is run. I'm considering reverting back to the original custom implementation and load only load `.env.test.local` or `.ent.test`.

 `@next/env` is actually able to merge multiple files, but I'd like to keep it simple. How do you see it?
 
Here is where  `@next/env` generates its dotfile list:
 ```js
  const isTest = process.env.NODE_ENV === 'test'
  const mode = isTest ? 'test' : dev ? 'development' : 'production'

  const dotenvFiles = [
    `.env.${mode}.local`,
    // Don't include `.env.local` for `test` environment
    // since normally you expect tests to produce the same
    // results for everyone
    mode !== 'test' && `.env.local`,
    `.env.${mode}`,
    '.env',
  ].filter(Boolean) as string[]
 ```

Meemaw: Hmm, I don't have enough context to confidently say what is better. However, I think we should in general aim to re-use `nextjs` parts where possible. And this seems to be a public API (we are not importing some `/dist/` folder modules)

toomuchdesign: Yep, it's a `next` dependency (but probably we should list it among ours deps). It's meant to be used by a real application (eg. it automatically loads env vars to global `process.env`, logs each time it's fired, etc), therefore it behaves in a way that sometimes clashes with our flow. Also: I'd like to reuse Next.js parts.

src/setEnvVars.ts
```diff
@@ -0,0 +1,83 @@
...
+const { SERVER, CLIENT } = RuntimeEnvironment;
+
+// @NOTE Next.js env var handling implementation is available here:
+// https://github.com/vercel/next.js/tree/v10.0.5/test/integration/env-config
+// We can't use it as long as this doesn't get fixed:
+// https://github.com/vercel/next.js/issues/21296
+let dotFile: Record<string, string> | undefined = undefined;
+export function loadDotFile({ nextRoot }: { nextRoot: string }) {
+  // https://github.com/vercel/next.js/blob/v10.0.5/packages/next-env/index.ts#L82
+  const dotFilePaths = [
```
Meemaw, line 16: Would it make sense to let user pass the `dotfile` path? I've noticed I'm loading a different file in my project (did it manually for now) `.env.development` https://github.com/insights-io/Insight/blob/master/frontend/app/src/__tests__/setup.ts#L7 and it would be annoying if I had to copy same contents of the file for `next-page-tester`  to load it.

toomuchdesign: This would also be a decent solution, even simpler than what we have now. Maybe a relative path from `nextRoot`?

If we exposed a `dotfilePath` option I'd probably get rid of default `dotFilePaths`. Just to not increase complexity exponentially. How do you see it?

Meemaw: Hmm, I think it still makes sense to fallback to those values. We can implement in a super simple way by just pushing the user `dotfilePath` in front.

```
if(dotfilePath) {
   dotFilePaths.unshift(dotfilePath)
|
```

toomuchdesign: I was considering edge cases like  having a `.env.test.local` in place (maybe for e2e tests) and preferring to run next-page-tester with no dotfiles at all (maybe declaring env var in-line).

A custom `dotfilePath` would cover any scenario at the cost of a bit more repetition. Otherwise I guess we would and up by forcing `dotfilePath` with a `false` boolean to disable dotfiles lookup.

Meemaw: If I understand correctly you are thinking about the option to disable `dotfile` loading.

We can make `dotfile?: string | false`:
- If nothing is passed; behaviour is as now
- If `false` is passed; dotfile loading is skipped
- If `string` is passed; path is added to 1st index of our `dotfilePath`

This seems to be flexible enough and covers all usecases?

About the repetition; Its up to user to handle it, but personally I would just wrap the `getPage` (and would encourage others to do so as well) to specify props used in all tests (e.g. `dotfile` , `useDocument`,...)


```js
export const getPage = (options) => {
   return getPageImpl({
       dotfile: '.env.development',
       useDocument: true,
       ...options,
     }
   )
}
```

toomuchdesign: Sorry @Meemaw, I was quite cryptic in my last comment. :)

Actually I was considering the opportunity to avoid introducing convoluted API's (like `dotfile?: string | false` + default loading behaviour) by dropping autoloading (what we do now) and only load the dotfile provided (if provided) via options.

Would it an acceptable solution?

Meemaw: Yeah, It works fine for me and it seems as a more "flexible" approach. We can always add the auto-loading back if we get a use-case/request/issue that makes sense.

PR #120, opened 2021-01-17
feat: support custom render page

Meemaw:

- feat: https://nextjs.org/docs/advanced-features/custom-document#customizing-renderpage
- feat: `nonIsolatedMoudles` option

☝️ both of these together fixes the initial document markup for CSS-in-JS libraries

Replaces some previous attempts: https://github.com/toomuchdesign/next-page-tester/pull/117, https://github.com/toomuchdesign/next-page-tester/pull/95 https://github.com/toomuchdesign/next-page-tester/pull/76

- https://nextjs.org/docs/advanced-features/custom-document#customizing-renderpage is not supported
- its not possible to specify `nonIsolatedModules` that should use same module instance between `server` and `client` environment. This is needed for cases where `React.Context` has to be used in `server` and `client` environment.

## What is the new behaviour?

- It is possible to specify pass `nonIsolatedModules` to `getPage` for advanced use-cases
- Custom "renderPage" is implemented which fixes the initial document render for CSS-in JS libraries

No


README.md
```diff
@@ -95,15 +95,16 @@ React element of the application.
...
+| **res**                        | Enhance default mocked [response object][res-docs]                                             | `res => res`       | -               |
+| **router**                     | Enhance default mocked [Next router object][next-docs-router]                                  | `router => router` | -               |
+| **useApp**                     | Render [custom App component][next-docs-custom-app]                                            | `boolean`          | `true`          |
+| **useDocument** (experimental) | Render [Document component][next-docs-custom-document]                                         | `boolean`          | `false`         |
+| **nextRoot**                   | Absolute path to Next.js root folder                                                           | `string`           | _auto detected_ |
+| **nonIsolatedModules**         | Array of modules that should use the same module instance in `server` and `client` environment | `string[]`         | -               |
```
Meemaw, line 107: This was needed to make `styletron-react` case properly work, but it should enable all the cases where external libraries are using `React.ContextProvider` on server and client.

Meemaw:
I plan to add some more isolated tests for the "nonIsolatedModules" and "custom render page" in case this gets positive feedback about being merged

src/commonTypes.ts
```diff
@@ -47,14 +53,17 @@ export type PageFile<FileType> = {
 
 export type PageParams = ParsedUrlQuery;
 
-export type PageObject = {
-  page: PageFile<NextPageFile>;
-  appFile: PageFile<NextAppFile>;
+export type RouteData = {
```
Meemaw, line 56: Splitting this from `PageObject`  is not required for this PR, but it will enable easier implementation of the custom `_error` page

src/getPage.tsx
```diff
@@ -53,83 +56,85 @@ export default async function getPage({
...
+        options={options}
+        routeData={pageObject}
+        makePage={(optionsOverrides) =>
+          makePage({ ...options, ...optionsOverrides })
+        }
+      >
+        {children}
+      </RouterProvider>
+    );
+  };
```
toomuchdesign, line 118: Could it be moved together with router related stuff?

src/getPage.tsx
```diff
@@ -53,83 +56,85 @@ export default async function getPage({
...
-      pageElement = (
+    const pageElement = (
+      <AppComponent Component={PageComponent} pageProps={pageData.props} />
+    );
+
+    if (
+      !useDocument ||
+      options.env !== RuntimeEnvironment.CLIENT ||
+      !headManager
+    ) {
```
toomuchdesign: I'm getting sleepy :) Why are we inverting conditional logic here?

toomuchdesign (commented):
Hi @Meemaw, massive PR :) A
I scattered around a few questions to help me better understand the details of this PR. :)

What is the scope of the newly introduced `InternalError`? When is it supposed to be called and how is different from default errors? Or it's only a way to centralise our error message creation?

Would you mind helping me understand how/where `RouteData` (`PageObject` subset) is used alone?

src/commonTypes.ts
```diff
@@ -25,16 +25,22 @@ export type Options = {
...
+  nonIsolatedModules?: string[];
 };
 
 export type OptionsWithDefaults = Required<Options>;
 
+export enum RuntimeEnvironment {
+  SERVER = 'server',
+  CLIENT = 'client',
+}
+
```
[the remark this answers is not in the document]
Meemaw: Moved

src/_app/getAppFile.ts
```diff
@@ -22,11 +22,14 @@ export function getAppFile({
       return customAppFile;
     }
   }
-  return getDefaultAppFile();
+  return getDefaultAppFile(nonIsolatedModules);
```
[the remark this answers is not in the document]
Meemaw: Done

src/makeRouterMock.ts
```diff
@@ -45,12 +45,12 @@ function makeDefaultRouterMock({
 }
 
 export default function makeRouterMock({
-  options,
-  pageObject: { pagePath, params, route, query },
+  routerEnhancer,
```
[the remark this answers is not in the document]
Meemaw: Changed back.

src/loadPage.ts
```diff
@@ -2,50 +2,50 @@ import path from 'path';
 import { existsSync } from 'fs';
 import { requireAsIfOnServer } from './server';
 import type { ExtendedOptions, PageFile } from './commonTypes';
+import { InternalError } from './_error/error';
 
 export function loadFile<FileType>({
   absolutePath,
+  nonIsolatedModules,
```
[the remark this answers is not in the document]
Meemaw: I would leave it as it is right now as I don't think that function will be requiring any more parameters, but if it does it should be ta trivial change to replace the parameter.

src/RouterProvider.tsx
```diff
@@ -1,39 +1,47 @@
...
 import { useMountedState } from './utils';
-import type { ExtendedOptions, Page, PageObject } from './commonTypes';
+import {
+  ExtendedOptions,
+  RuntimeEnvironment,
+  RouteData,
+  MakePageResult,
+} from './commonTypes';
+
+type Props = {
```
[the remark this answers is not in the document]
Meemaw: Personally code feels less cluttered to me if type definitions are not inlined and diffs are smaller when type needs to be reused/exported. It's just a preference, I can inline it if you prefer that style.

src/makePageElement.ts
```diff
@@ -1,41 +1,49 @@
...
+}): PageConstructs {
+  const AppComponent = pageObject.appFile[env].default;
+  const PageComponent = pageObject.page[env].default;
 
-  return { pageElement, pageObject, pageData };
+  return {
+    AppComponent,
+    PageComponent,
+    routeData: pageObject,
+  };
```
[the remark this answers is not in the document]
Meemaw: Doesn't really do much. It seems I renamed it shortly after you started reviewing. Its called "getPageComponents" now and it just returns `"AppComponent"` and `"PageComponent"` for correct `"env"` from the nested objects.

src/makePageElement.ts
```diff
@@ -1,41 +1,49 @@
...
+  RuntimeEnvironment,
+} from './commonTypes';
 
 /*
- * Return an instance of the page element corresponding
- * to the given path
+ * Return page info associated with a given path
  */
-export default async function makePageElement({
+export async function getPageInfo({
```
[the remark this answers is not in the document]
Meemaw: Re-introduced it to simplify a few things. Let me know if it makes sense now.

src/getPage.tsx
Meemaw: Its hard to move it in current shape as its a closure over a few objects available in scope: `makePage`, `pageObject`, ...

It could be moved if we passed all those to it as arguments.

src/getPage.tsx
Meemaw: I think I changed that back. Probably shortly after you started you review.

Meemaw:
> Hi @Meemaw, massive PR :) A
> I scattered around a few questions to help me better understand the details of this PR. :)

Yeah, sorry about the massive PR. I probably touched more things than I should and laid grounds for the custom `_error` page.

> 
> What is the scope of the newly introduced `InternalError`? When is it supposed to be called and how is different from default errors? Or it's only a way to centralise our error message creation?

There are 2 reasons for introducing it:
1. Centralise message creation
2. With custom `_error` page we have to differentiate between errors thrown in user code (to render custom `_error` page) and errors thrown from our code (for better DX). By extending the `Error` we can easily check if the error was thrown by us and thus escape it in our catch handler

> 
> Would you mind helping me understand how/where `RouteData` (`PageObject` subset) is used alone?

(renamed it to `RouteInfo`).

It is the case yet in this PR, but I expect it will become clear with custom `_error` page implementation.

For example if you look at our `getPageObject` you can see it returns `PageObject`. However if the route is not found it currently throws an error (e.g. `page` and `appFile` do not exist in that case). With the custom `_error`, we will at this point  throw a `404Error` and we would want to attach `RouteInfo` to it as it will be required somewhere upstream to render the `_error` / `404` page. We have the `RouteInfo` from the `getRouteInfo` call (1st line of the function) so we can easily do that (e.g. attach it to our custom `Error` object that will have to be introduced.

Let me know if this makes sense. If not I think we can delay that change (it should be trivial to revert it) until it is absolutely needed.

src/commonTypes.ts
```diff
@@ -47,17 +49,20 @@ export type PageFile<FileType> = {
 
 export type PageParams = ParsedUrlQuery;
 
-export type PageObject = {
-  page: PageFile<NextPageFile>;
-  appFile: PageFile<NextAppFile>;
+export type RouteInfo = {
```
Meemaw, line 52: This is matching the return type of `getRouteInfo`: https://github.com/toomuchdesign/next-page-tester/pull/120/files#diff-768042b7438a04ad39b577c36622fd2716eec2cd53c5d896700b4383e055acceR70

Meemaw:
@toomuchdesign I've refactored the PR to reduce diff and address most of the comments. I think the changes should be much easier to review now. Let me know what you think.

Also added some more tests to get test coverage back to 100% 🎉

src/loadPage.ts
```diff
@@ -2,50 +2,53 @@ import path from 'path';
...
 
-export function loadPageIfExists<FileType>({
-  pagePath,
-  options,
-}: {
-  pagePath: string;
-  options: ExtendedOptions;
-}): PageFile<FileType> | undefined {
+export function loadPageIfExists<FileType>(
+  options: LoadPageOptions
```
toomuchdesign, line 48: I've stumbled twice upon this `options` types which sounds like our `options` object but it's not. On one side we avoid repetition, on the other types are less explicit. We'll see further on... :)

src/utils.ts
```diff
@@ -90,14 +89,23 @@ export function useMountedState(): () => boolean {
...
   'next/router',
   'next/dist/next-server/lib/head-manager-context',
   'next/dist/next-server/lib/router-context',
   'next/dist/next-server/lib/runtime-config',
 ];
-export function executeWithFreshModules<T>(f: () => T): T {
+
+export function executeWithFreshModules<T>(
+  f: () => T,
+  userNonIsolatedModules: string[]
```
toomuchdesign, line 102: Would it improve readability?

```suggestion
  options: {userNonIsolatedModules: string[]}
```

src/makePageElement.ts
```diff
@@ -1,41 +1,61 @@
...
 import { renderApp } from './_app';
-import type { ExtendedOptions, Page } from './commonTypes';
 
 /*
- * Return an instance of the page element corresponding
- * to the given path
+ * Return page info associated with a given path
  */
-export default async function makePageElement({
+export async function getPageInfo({
```
toomuchdesign, line 15: I don't remember... does it need to be exported?

toomuchdesign (changes requested):
I loved  `custom-render-page` tests. 👍 

There's something to rephrase in the readme and not much more. Document rendering is becoming a sort of project in the project, let's hope it won't require too much maintenance effort.

Thanks so much for this wild ride @Meemaw  :)

src/makePageElement.ts
Meemaw: Yes, its used in `getPage`.

src/utils.ts
Meemaw: Changed to object.

README.md
```diff
@@ -175,10 +176,16 @@ Note: `document.cookie` does not get cleaned up automatically. You'll have to cl
 
 Next.js `Link` component invokes `window.scrollTo` on click which is not implemented in JSDOM environment. In order to fix the error you should [set up your test environment](#set-up-your-test-environment) or provide [your own `window.scrollTo` mock](https://qiita.com/akameco/items/0edfdae02507204b24c8).
 
+### Warning: Text content did not match. Server: "x" Client: "y"" error
+
+This warning can appear during client side hydration of the page and can be due to a bug in your code or due to to the fact that `next-page-tester` has to isolate modules in `server` and `client` environment to avoid "environment leaking". In second case, modules that needs to use same instance in `server` and `client` environment (e.g. `React.Context`, or `css-in-js` libraries) can be marked as such by being passed in the `nonIsolatedModules` option of the `getPage` function; [styletron-react example](src/__tests__/third-party/styletron-react/styletron-react.test.ts#L13)
```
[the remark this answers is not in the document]
Meemaw: It makes sense. I know in actual SSR application things work differently, but if I understand our current approach correctly, we need to preserve module identity for modules that are using "React" constucts to avoid issues.

Meemaw:
> I loved `custom-render-page` tests. 👍
> 
> There's something to rephrase in the readme and not much more. Document rendering is becoming a sort of project in the project, let's hope it won't require too much maintenance effort.
> 
> Thanks so much for this wild ride @Meemaw :)

Well, I think `_document` rendering is the most complex part of `Next.JS` (and hardest to replicate). I think we are on a good way to make `non-experimental` (it doesn't implement everything yet, but it seems stable lately).


Related, but for different PR. During some test writing yesterday I noticed we are not populating `document.title` (and some others if I'm not mistaken). It would make DX better for testing it `expect(document.title).toEqual("Homepage")`

README.md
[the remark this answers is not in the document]
toomuchdesign: The fact that we use the same module identity between server and client renders it's actually a quite nasty workaround, which I expect leading to some bug sooner or later.

What we'd need (and we don't have yet) is preserving imported module identity for a whole rendering cycle in a given environment: 1 for server render, a new one for client render, and a new one for every subsequent data fetching call.

Since we don't have parallel separated environment we switch between server and client: client is currently the same natural scope where tests run, while we try to enclose server operation in `executeAsIfOnServer` and `requireAsIfOnServer ` scopes.

Since each call to `requireAsIfOnServer` generates fresh modules, it's here where we break code which relies on module identity. Therefore we whitelist modules which identity we want to preserve throughout the whole session, with the side effect of letting their identity leaking into `client` execution (which should have its own modules instances).

Last point: `Text content did not match` could be due to a bug in userland code but it just signal that server and client render output are different.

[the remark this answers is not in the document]
Meemaw: Quick though that passed my mind and maybe an option to explore one day.

What if we did SSR in a separate process (we would spawn a new process without JSDOM environment) and just do the server render there and we would get the "html" from it. Everything that happens in the main process is "client" environment. 

If this works conceptually, we would get rid of all these problems.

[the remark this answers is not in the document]
toomuchdesign: I also wondered whether a similar approach could be feasible. Since I've never manually spawned Node.js processes I've never tried out. I wonder whether it could have downsides in terms of handling errors thrown in spawn processes. I'd love to explore this direction.

Also, there is another simpler option which which I still haven't investigated (but I suspect might fall short of flexibility): which is keeping our rendering phases linear and avoid mixing server and client logic. We emulate sever render first and client render afterwards.

Meemaw:
@toomuchdesign any blockers for this one? There were many discussions in here so just want to make sure I didn't forget any.

README.md
```diff
@@ -175,10 +176,18 @@ Note: `document.cookie` does not get cleaned up automatically. You'll have to cl
 
 Next.js `Link` component invokes `window.scrollTo` on click which is not implemented in JSDOM environment. In order to fix the error you should [set up your test environment](#set-up-your-test-environment) or provide [your own `window.scrollTo` mock](https://qiita.com/akameco/items/0edfdae02507204b24c8).
 
+### Warning: Text content did not match. Server: "x" Client: "y"" error
+
+This warning means that your page renders differently between server and browser. This can be an expected behavior or signal a bug in your code.
+
+The same error could also be triggered when you import modules that, in order to work, need to preserve module identity between `server` and `client` environment (e.g. `React.Context` or `css-in-js` libraries). In this case you can list these modules in `nonIsolatedModules` option to preserve their identit: [see styletron-react example](src/__tests__/third-party/styletron-react/styletron-react.test.ts#L13)
```
toomuchdesign, line 183:
```suggestion
The same error could also be triggered when you import modules that, in order to work, need to preserve module identity throughout the entire SSR cycle (from the moment they are imported to SSR rendering): e.g. `React.Context` or `css-in-js` libraries. In this case you can list these modules in `nonIsolatedModules` option to preserve their identity: [see styletron-react example](src/__tests__/third-party/styletron-react/styletron-react.test.ts#L13)
```

toomuchdesign (approved):

Issue #121, opened 2021-01-19
routerProvider push does not support next's URL object (with pathname, query etc) format.


DavidOrchard:
## 🐛 Bug Report

Calling next's router.push with url object containing pathname and query results in [Object] object because routerProvider push mock does url.toString() instead of the more correct "next/dist/next-server/lib/utils".formatWithValidation call.

## To Reproduce

Call router.push({pathname: "foo", query: {bar: "baz"}});

## Expected behavior

the mock push is called with nextRoute("/foo?bar=baz");

## Run `npx envinfo --preset jest`

Paste the results here:

```bash
  System:
    OS: macOS 10.15.6
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
  Binaries:
    Node: 12.16.0 - ~/.nvm/versions/node/v12.16.0/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.16.0/bin/npm
  npmPackages:
    jest: 26.6.1 => 26.6.1 
```

Meemaw:
Hey @DavidOrchard, thanks for reporting this. I've reproduced the issue and prepared a fix https://github.com/toomuchdesign/next-page-tester/pull/122

PR #122, opened 2021-01-19
fix: client side object navigation

Meemaw:

Fixes https://github.com/toomuchdesign/next-page-tester/issues/121

Client side navigation with `UrlObject` does not work

## What is the new behaviour?

Client side navigation with `UrlObject` works

No


toomuchdesign (approved):
Laser cut PR 👍

PR #130, opened 2021-01-22
refactor: introduce dotenvFile option

toomuchdesign:

Refactor

dotenv file lookup happens automatically against a given list of paths.

## What is the new behaviour?

Remove automatic lookup and provide a `dotenvFile` option to explicitly load a dotfile.

Yes. But previous behaviour hasn't been released, yet.


src/setEnvVars.ts
```diff
@@ -4,30 +4,34 @@ import dotenv from 'dotenv';
...
-    }
+  const dotenvFilePath = path.resolve(nextRoot, dotenvFileRelativePath);
+  if (existsSync(dotenvFilePath)) {
+    dotenvFile = dotenvExpand({
+      parsed: dotenv.parse(readFileSync(dotenvFilePath)),
+      // @ts-expect-error dotenv-expand type definition is out of date
+      ignoreProcessEnv: true,
+    }).parsed;
+  } else {
+    throw new InternalError(`Cannot find env file at path: ${dotenvFilePath}`);
```
Meemaw, line 34: Wondering if we should throw or perhaps just log a `warning`? WDYT?

toomuchdesign: Yep, Why not. :)

Meemaw (approved):

PR #131, opened 2021-01-22
feat: render 404 when matching route not found

Meemaw:
### Feature

- render `404` page when matching route not found

When matching route is not found an error is thrown.

## What is the new behaviour?

When matching route is not found `404`  page is rendered

Might be, depending on how your application was tested.


src/__tests__/404/404.test.ts
```diff
@@ -0,0 +1,62 @@
+import { getPage } from '../../../src';
+import path from 'path';
+import { screen } from '@testing-library/react';
+import { stripReactExtraMarkup } from '../__utils__/expectDOMElementsToMatch';
+
+describe('404', () => {
+  describe.each([
+    ['ssr "notFound" return', '/a'],
+    // ['matching page not found', '/random'],
```
Meemaw, line 9: Tests stayed the same, but were wrapped in top level `404` describe so they can be reused. When implemented, the lined with `'matching page not found'` can be un-commented and tests should pass

Meemaw:
@toomuchdesign this PR doesn't implement anything yet, as I would like to discuss how to implement this nicely. It might seem trivial, e.g. just replace the `throw new InternalError('No matching page found for given route');` inside `getPageObject` function, but it is not that easy.

I've already mentioned this problem in the initial custom `_error` PR. If we are not finding a match for `page` we cannot construct `PageObject` type. This is a problem, because all the functions upstream are expecting this object. One way would be to populate it as much as possible and fake the rest (or see how `next.js` populates these fields).

toomuchdesign:
Offhand I'd wander what page object I expect from a non-matching route.

If I don't overlook edge cases, it seems to me that since currently `getPageObject` returns a `PageObject` or an error, the new behaviour could consist in returning a normal `PageObject` or a special `NotFoundPageObject`. Are there other cases to be handled?

Now the 2nd question is, what would `NotFoundPageObject` contain?

I see that the `router` object for a custom 404 page is:

```js
asPath: "/non-matching-route"
back: ƒ ()
basePath: ""
beforePopState: ƒ ()
components: {/404: {…}, /_app: {…}}
defaultLocale: undefined
events: {on: ƒ, off: ƒ, emit: ƒ}
isFallback: false
isReady: true
locale: undefined
locales: undefined
pathname: "/404"
prefetch: ƒ ()
push: ƒ ()
query: {}
reload: ƒ ()
replace: ƒ ()
route: "/404"
```
Do you see any obstacle in building a `NotFoundPageObject` with the same signature as `PageObject`?

Or maybe you're thinking of different approaches?

Meemaw:
@toomuchdesign fell free to commit any refactors you had in mind into this PR.

toomuchdesign:
I'm considering adding a test to ensure `NotFoundPageObject` has the expected values. I'll write it:) We have a few options:
- testing a custom 404 in router-mocker
- testing a custom 404 somewhere in page-data-fetching
- testing router or data fetching in dedicated 404 test folder

What do you think is the best options @Meemaw?

That said, I guess time has come to tidy up our src folder. We can do this after we merge this PR. You welcome to drop any suggestion.

Have a bright day!

Meemaw:
> I'm considering adding a test to ensure `NotFoundPageObject` has the expected values. I'll write it:) We have a few options:
> 
> * testing a custom 404 in router-mocker
> * testing a custom 404 somewhere in page-data-fetching
> * testing router or data fetching in dedicated 404 test folder
> 
> What do you think is the best options @Meemaw?

I think it makes sense to add it to the dedicated `404` folder.

toomuchdesign (approved):
We're all done here :)
We can merge and maybe spend a few keystrokes to tidy up things, unless we need to publish immediately.

Meemaw:
> We're all done here :)
> We can merge and maybe spend a few keystrokes to tidy up things, unless we need to publish immediately.

I think we can merge this, but there is actually 1 more thing I would like to add thats related to 404 pages.

In my project I have some SSR redirects to external URLS. That looks something like this:

```js
return {
   redirect: {
     destination: 'https://www.example.com',
     permanent: true,
  }
}
```

With the current implementation this `route` of-course wont be found and `404` page will be rendered. My suggestion is to check if the `URL` is absolute and in that case do something differently to make it testable. I think easiest solution would be to just throw an error like this:

```js
throw new InternalError(`External route navigation: ${route}`)
```

☝️ this could easily be testable in the user code.

toomuchdesign:
Yep let's throw error 👍 
I think throwing is quite a powerful pattern which is very suitable to not only signal errors.

Meemaw:
Okay, I'll merge this and prepare a separate PR for it -- to make changes easier to review.

PR #135, opened 2021-01-25
feat: implement 404 when no matching page found

Meemaw:

## What is the new behaviour?

...


Meemaw:
@toomuchdesign this is a proposed implementation for the 404 when matching page is not found. Its targeting another PR: https://github.com/toomuchdesign/next-page-tester/pull/131, thus the omitted PR description. If this approach seems fine, I will add some additional test cases to make sure we are correctly matching/constructing the http request parameters (pathname) as next.js does.

src/getPageObject.ts
```diff
@@ -31,10 +34,25 @@ export default async function getPageObject({
...
+    pagePath: pathname,
+    params: {},
+    paramsNumber: Object.keys(query).length,
+    query,
+    resolvedUrl:
+      pathname +
+      stringifyQueryString({ object: query, leadingQuestionMark: true }),
+    route,
+  };
 }
```
toomuchdesign, line 56: Let's refactor `getPageObject.ts` at the end if this development cycle. I think that with the right abstraction we might avoid some duplication. Good so far.

src/makePageElement.ts
```diff
@@ -19,8 +19,11 @@ export async function getPageInfo({
   options: ExtendedOptions;
 }): Promise<PageInfo> {
   const pageObject = await getPageObject({ options });
-  let pageData = await fetchRouteData({ pageObject, options });
+  if (!pageObject.page) {
```
toomuchdesign, line 22: Shall we add a sort of special property/flag to explicitly signal the pageObject type?

src/_error/getErrorFile.ts
```diff
@@ -15,14 +15,6 @@ export function getErrorFile({
...
 
   if (customErrorFile) {
-    // Assign "origGetInitialProps" to custom _error page if needed
-    if (!customErrorFile.client.default.getInitialProps) {
-      const defaultErrorFile = getDefaultErrorFile({ nonIsolatedModules });
-      customErrorFile.client.default.getInitialProps =
-        defaultErrorFile.client.default.origGetInitialProps;
-      customErrorFile.server.default.getInitialProps =
-        defaultErrorFile.server.default.origGetInitialProps;
-    }
```
toomuchdesign, line 25: When have we introduced this? It looks totally new to me 😄

toomuchdesign (commented):
Hi @Meemaw. It looks a quite linear implementation to me. 👍 

There just one detail I'd like to hear you ideas about:

I was wandering whether we should avoid the distinction between ` NotFoundPageObject | PageObject` and try to return from `getPageObject` with a valid `PageObject` even for a 404.

We might maybe add a `type` prop to `PageObject` to tell a normal page from a 404 instead of duck typing `page` prop existence.

src/__tests__/routing/routing-static-routes.test.tsx
```diff
@@ -18,14 +19,16 @@ describe('Static routes', () => {
...
+      const { render } = await getPage({
+        nextRoot,
+        route: '/blog/5/doesntexists',
+      });
+      render();
+      expect(screen.getByText('404')).toBeInTheDocument();
+
+      expect(
+        screen.getByText('This page could not be found.')
+      ).toBeInTheDocument();
```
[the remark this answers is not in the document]
Meemaw: Done

src/_error/getErrorFile.ts
Meemaw: It was introduced in previous PR with initial 404 implementation. When custom error page has no `getInitialProps`, we should still provide it with the default props.

However, its not smart to patch "cached" modules like this (it leaks into other tests).

src/makePageElement.ts
Meemaw: Current approach takes advantage of Typescript's [discriminating-unions](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) over the `page` property to correctly handle the cases when `page` is available and when it is not.

If we try to reuse the same 'PageObject` with the `page` defined we would probably need to use some `@ts-ignore` and "lie" to the compiler which I usually try to avoid.

What exactly did you have in mind here? Fell free to open a PR targeting this branch with your suggestion 🚀

src/getPageObject.ts
Meemaw: Lets keep refactoring our of this PR to keep the diff smaller and do a separate PR after this is merged and before we merge into `master`. WDYT?

toomuchdesign: Yo! :)

src/makePageElement.ts
toomuchdesign: I didn't have in mind any TS trickery, TS is already tricky enough!

I wondered whether the way we currently tell `NotFoundPageObject` from `PageObject` by checking `page` prop could be more explicit.

Do you see any improvement if the discrimination between `NotFoundPageObject` and `PageObject` was something like a `type` prop instead of the presence of `page` prop?

Meemaw: Do you think current check is not explicit enough?

I think that `if (!pageObject.page)` clearly indicates that there is no page found, but perhaps that just for me.

Were you thinking about something along these lines: `if(pageObject.type === 'notFound')`?

Meemaw: Do you think current check is not explicit enough?

I think that `if (!pageObject.page)` clearly indicates that there is no page found, but perhaps that just for me.

Were you thinking about something along these lines: `if(pageObject.type === 'notFound')`?

Meemaw: Do you think current check is not explicit enough?

I think that `if (!pageObject.page)` clearly indicates that there is no page found, but perhaps that's just me.

Were you thinking about something along these lines: `if(pageObject.type === 'notFound')`?

toomuchdesign: Yep! It's not a game changer but I was wondering if setting an explicit type prop would make sense.

[Discriminating unions docs](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) also use the same strategy.

Meemaw: Added in https://github.com/toomuchdesign/next-page-tester/pull/135/commits/f22b2c06611af8f7eda206ea7569008807d3872a

toomuchdesign:
Are we going to need a `pageObject` to render error pages, too?

Meemaw:
> Are we going to need a `pageObject` to render error pages, too?

Probably, but I don't know the details yet. To implement error pages we will have to catch them at multiple places so its hard to say:
- data fetching methods
- initial render
- client side navigation
- client side render (not sure about this one actually -- would have to check nextjs)

Meemaw:
Any blockers to merge this @toomuchdesign ? 

P.S. this is not targeting master so fell free to do any refactoring you had in mind after this one is merged.

toomuchdesign:
Let's merge!

Issue #142, opened 2021-02-03
styled-components' context doesn't work


Doaa-Ismael:
## 💬 Questions and Help
When I wrap ```<Component />``` with a Provider ```<ThemeProvider theme={theme} >``` in _app and then try to consume this theme or provider value it returns undefined instead of the provider value.

Please help, I have been 2 days debuging to find the issue and there is no luck

home.test.jsx
```js
it('should renders home page', async () => {
    const { render} = await getPage({
         route: '/home',
          useApp: true
    });
    render();
    expect(screen.getByText('home'))
    expect(screen.getByText('child'))
});
```

_app.js

```js
import React, {useContext} from 'react';
import styled, { ThemeProvider, ThemeContext } from 'styled-components';
import ThemJson from "../theme";

const StyledContainer = styled.div`
    background-color: ${props => { 
        console.log("PROPS IN STYLED COMPONENTS", props.theme);  // output is correct
        return props.theme.color.primary;
    }
};
`
const AppChild = () => {
    const theme = useContext(ThemeContext);
    console.log(theme) // Output correct
    return (<StyledContainer>home</StyledContainer>);
};

const MyApp = ({ Component }) => {
  return (
      <ThemeProvider theme={ThemJson}>
         <AppChild />
         <Component />
    </ThemeProvider>
  );
};
export default MyApp
```

home.jsx
```js
import React, {useContext} from 'react';
import styled, {ThemeContext} from 'styled-components';

export const StyledContainer = styled.div`
    background-color: ${props => {
        console.log("PROPS IN STYLED COMPONENTS", props.theme); // ERROR can't red property theme of undefined
        return props.theme.color.primary;
    }
};
`

function Home(props) {
    const theme = useContext(ThemeContext);
    console.log(theme) // Wrong!! // theme is undefined
    return (
        <StyledContainer>
            child
        </StyledContainer>
    );
};

export default Home;
```

theme.json
```json
{
    "color": {
        "primary": "red"
    }
}
```

Meemaw:
Hey @Doaa-Ismael 👋 

Could you try adding the following option to `getPage` call:
`nonIsolatedModules: ['styled-components']`

Let me know if that doesn't work for you

Doaa-Ismael:
> Hey @Doaa-Ismael 👋
> 
> Could you try adding the following option to `getPage` call:
> 
> ```js
> nonIsolatedModules: ['styled-components']
> ```
> 
[...]

PR #143, opened 2021-02-07
Perform SSR in one go

toomuchdesign:

Refactor/Bug fix
 
Rendering involves several switches between server and client environment, forcing the introduction of `nonIsolatedModules` option which provides a workaround to allow code that relies on module identity to work.

This approach has limitations: #142 

## What is the new behaviour?

Try to perform all the SSR related operation in one single go, to avoid client/server switching and hopefully make `nonIsolatedModules` option not necessary anymore.

In this PR we replaced original `PageObject`'s `page` and `appFile` props with a single `files` object with the following shape:

```ts
type files = {
  client: {
    documentFile: NextDocumentFile;
    appFile: NextAppFile;
    pageFile: NextErrorFile | NextPageFile;
  };
  server: {
    documentFile: NextDocumentFile;
    appFile: NextAppFile;
    pageFile: NextErrorFile | NextPageFile;
  };
};
```

The main change is the way we load the files necessary fo render: instead of loading each file in client/render env, we now have to load all the required files in a single environment switch, so that we preserve modules identity between `documentFile`, `appFile`, `pageFile` and everything imported from them.

These 3 files are currently all we need to render (and data fetch) a Next page. If Next architecture changed and we needed 1000 files, we would have to import these 1000 files in one single go. File loading logic mainly happens [here](https://github.com/toomuchdesign/next-page-tester/blob/ssr-one-go/src/getNextFiles.ts).

I was also able to eventually remove `noIsolatedModules` option, even though we still need to non-isolate those modules which we import within our library code. Maybe we'll be able to find a workaround.

As you can see in CommonTypes, I had to explicitly declare 2 different `NextFiles` (`NextPageFiles` | `NextErrorPageFiles`) and `PageObject` (`PageObject` | `NotFoundPageObject`). This allowed to have a more linear typing flow, even though this brings in some duplication.

@Meemaw  Happy to hear your thoughts and to provide clarifications 😅 

PS.
I had to strictly separate `PageObject` creation in case of found and not found page. When we realise we have a not found page, we now forget about the previously created PageObject and we [create a fresh one](https://github.com/toomuchdesign/next-page-tester/blob/ssr-one-go/src/404/makeNotFoundPageObject.ts).

No


toomuchdesign:
After spending a few time experimenting with this PR I got to the conclusion that in order to find a solution for #142 we should avoid letting users use `nonIsolatedModules` option. In case of `styled-components` `nonIsolatedModules` fixes module identity between import and render time but breaks modules initialisation environment.

Therefore this PR will explore a way to SSRing in a way that does not conflict with browser modules initialization and that could  make such option no more necessary. 👀

Meemaw:
> In case of `styled-components` `nonIsolatedModules` fixes module identity between import and render time but breaks modules initialisation environment.

Well said 👆 I knew this was the issue but stuggled to put it in words 😅

Meemaw:
Hey @toomuchdesign whats the status on this? I'm super excited if we get this to work 🤩 It would almost justify putting `useDocument` out of experimental.

toomuchdesign:
Hi @Meemaw! I'm so scared by what's inside `_document` folder 😄 .

In regard of this PR, it's currently surprisingly working. It needs quite a bit of refactoring since I duplicated most of the logic related to loading files in page, app. But the strategy of loading all the SSR related files in a single go, seems to work.

Beside that I guess we might refactor `documentObject` structure and the way we preload client and server file sets.

Meemaw:
Nice.

Have you also tried the styled-components reproducer?

toomuchdesign:
No, I haven't. But I tried to abstract that scenario in:
https://github.com/toomuchdesign/next-page-tester/tree/ssr-one-go/src/__tests__/modules-initialization-env

toomuchdesign:
Updates: I'm going to merge but I don't know when.
This going to be a quite deep refactor of the data flows and I can't predict how long it's going to take and how much time I'm going to be able to spend on this. Btw, I'm actively working on it in my spare time.

src/__tests__/modules-initialization-env/modules-initialization-env.test.ts
```diff
@@ -0,0 +1,32 @@
+import path from 'path';
+import { screen } from '@testing-library/react';
+import { getPage } from '../../../src';
+import { silenceConsoleError } from '../__utils__';
+
+silenceConsoleError('Text content did not match.');
```
Meemaw, line 6: Is it expected behaviour that text content does not match? If so I think we should assert this and not just silence the error.

src/_app/render.tsx
```diff
@@ -3,25 +3,24 @@ import React from 'react';
...
-import { getPageComponents } from '../makePageElement';
 
 export default function renderApp({
   options,
   pageObject,
   pageProps,
 }: {
   options: ExtendedOptions;
-  pageObject: PageObject;
+  pageObject: GenericPageObject;
```
Meemaw, line 16: Nit: having `Generic` in type name seems a bit redundant as `PageObject` seems generic enough. Probably means the "found" case should have a more detailed name.

src/commonTypes.ts
```diff
@@ -51,35 +50,34 @@ export type ExtendedOptions = OptionsWithDefaults & {
...
 };
 
 export type NotFoundPageObject = RouteInfo & {
   type: 'notFound';
-  appFile: PageFile<NextAppFile>;
+  absolutePagePath: string;
+  files: MultiEnv<NextErrorPageFiles>;
 };
 
+export type GenericPageObject = PageObject | NotFoundPageObject;
```
Meemaw, line 79: Just a suggestion. Actually don't have good ideas:

```suggestion
export type GenericPageObject = ExistingPageObject | NotFoundPageObject;
```

src/commonTypes.ts
```diff
@@ -119,17 +122,28 @@ export class CustomError extends Error {
...
-export type PageInfo = {
-  pageObject: PageObject;
-  pageData: PageData;
+export type NextErrorPageFiles = {
+  documentFile: NextDocumentFile;
+  appFile: NextAppFile;
+  pageFile: NextErrorFile;
+};
+
+export type NextGenericPageFiles = NextPageFiles | NextErrorPageFiles;
```
Meemaw, line 138: Same, just a suggestion.

```suggestion
export type NextPageFiles = NextExistingPageFiles | NextErrorPageFiles;
```

src/getNextFiles.ts
```diff
@@ -0,0 +1,79 @@
...
+import { executeAsIfOnServerSync } from './server';
+import type {
+  ExtendedOptions,
+  NextPageFiles,
+  NextErrorPageFiles,
+  MultiEnv,
+} from './commonTypes';
+
+// Get Document, App and Page files
+// @TODO Find a way to avoid this duplication and let TS follow pageFile typing
```
Meemaw, line 14: Cant you use a generic function here?

Meemaw (approved):
Generally LGTM!

Left a few minor comments!

toomuchdesign (commented):
Hi @Meemaw,
renamed a few things around and left a couple comments :)

Meemaw:
> Yep, I'd say it's expected. Since the test didn't focus on React triggering the error, I went for silencing that specific error for > the sake of simplicity and to not clutter the console.
> 
> Shall we consider rewriting silenceConsoleError utility as something that asserts against global console.error being invoked > with an expected payload?

Not a big deal, just feels a bit off to silence `console.errors`. As I see it there are 2 cases:
- Error is expected; we should assert it happens
- Error is not expected; it is a bug in our code and we should fix it in the future but there is no time ATM; we should still add an assertion it happens and flip the condition to make sure it doesn't happen when we fix it.

Regardless, fell free to merge as I'm sure a lot of people are excited about this release (myself included) 🙂

toomuchdesign:
Ups, it seems I overlooked your message when I merged.

I'd say `silenceConsoleError` was used when error log was expected but didn't matter and wasn't neither the focus of the test. Eg: testing `render` method rendering a value which depends on client/server environment.

`silenceConsoleError` is meant to silence specific error messages.

Let's address it in another PR 👍

Issue #151, opened 2021-02-17
env vars aren't available within next.config.js


jesstelford:
## 🐛 Bug Report

Next.js will load env vars before processing `next.config.js`, allowing access of `process.env.WHATEVER` to inform config values. `next-page-tester` does not load env vars until _after_ `next.config.js` is processed, resulting in `process.env.WHATEVER` always being `undefined`.

## To Reproduce

Create a `.env.local` with:

```
DOMAIN=https://example.com
```

Create a `next.config.js` with:

```javascript
module.exports = {
  i18n: {
    locales: ['en-AU'],
    defaultLocale: 'en-AU',
    domains: [
      {
        domain: process.env.DOMAIN,
        defaultLocale: 'en-AU',
      },
    ],
  },
}
```

Then run any test. Note Next throws an error:

```
    Invalid i18n.domains values:
    {"defaultLocale":"en-AU"}

    domains value must follow format { domain: 'example.fr', defaultLocale: 'fr', locales: ['fr'] }.
    See more info here: https://err.sh/next.js/invalid-i18n-config
```

The object is missing the `"domain"` key (because `process.env.DOMAIN` is `undefined`)

## Expected behavior

The env vars should be loaded before loading the config, and subsequently Next.js should not throw an error here.

## Link to repl or repo

Minimal reproduction: https://github.com/jesstelford/next-page-tester-bug-151

## Run `npx envinfo --preset jest`

Paste the results here:

```bash
  System:
    OS: macOS 10.15.7
    CPU: (4) x64 Intel(R) Core(TM) i5-7267U CPU @ 3.10GHz
  Binaries:
    Node: 12.20.1 - ~/.nvm/versions/node/v12.20.1/bin/node
    Yarn: 1.19.1 - ~/.config/yarn/global/node_modules/.bin/yarn
    npm: 6.14.9 - ~/dev/test/node_modules/.bin/npm
  npmPackages:
    jest: ^26.6.3 => 26.6.3
```

Meemaw:
@jesstelford thanks for reporting this. Its indeed a bug in how we handle env/config variables.

I've prepared a [fix](https://github.com/toomuchdesign/next-page-tester/pull/152) that should solve your issues.

PR #152, opened 2021-02-17
fix: next config/env loading

Meemaw:

Fixes the way (order) in which we load environment variables.

Fixes https://github.com/toomuchdesign/next-page-tester/issues/151

We load next config before we load base environment variables. This was due to the reason that when we called `setEnvVars` we expected next config to be loaded already. So we had a dependency cycle in which we couldn't  set env vars before loading next config.

## What is the new behaviour?

We load base environment variables (dotenv + runtime) before loading next config. After that we have all the variables available for the `next.config.js` to load properly.

After, in `setEnvVars` we just merge in the additional env vars coming from `next.config.js` itself.

No


src/__tests__/environment-variables/environment-variables.test.tsx
```diff
@@ -36,52 +39,62 @@ describe('Environment variables', () => {
...
-            FROM_RUNTIME: 'FROM_RUNTIME',
-            FROM_CONFIG: 'FROM_CONFIG',
-            FROM_DOTFILE: undefined,
-            FROM_DOTFILE_EXPANDED: undefined,
+            NEXT_PUBLIC_FROM_RUNTIME: 'NEXT_PUBLIC_FROM_RUNTIME',
             NEXT_PUBLIC_FROM_DOTFILE: 'NEXT_PUBLIC_FROM_DOTFILE',
-
-            NAME_CLASH_RUNTIME_VS_CONFIG: 'FROM_CONFIG',
-            NAME_CLASH_RUNTIME_VS_DOTFILE: 'FROM_RUNTIME',
-            NAME_CLASH_RUNTIME_VS_CONFIG_VS_DOTFILE: 'FROM_CONFIG',
```
Meemaw, line 60: Looking at this test it seems our env loading didn't work properly before. We are doing client `render` and we got all the variables into the DOM (including those not prefixed with `NEXT_PUBLIC_`)

toomuchdesign (changes requested):
Hi @Meemaw,
I'm submitting a request change because I feel like the way I understood environment variables in Next.js are different then what I see here and I'd like to clarify my assumptions with you.

First, I really like the way you rearranged env var tests 👍 

My point concerns client runtime. Server runtime seems ok to me. Maybe I'll double check it later.

In regard of client runtime I though that the `NEXT_PUBLIC_` prefix filtering happened only for those env vars declared in dotenv files. I would expect any other var coming from runtime or config file to be available also in client runtime.

I've done a quick check and It seems to work like this to me. Would mind double checking and help me clarify this point?

Maybe we could rename client runtime as something like public runtime or browser runtime?

Meemaw:
Hey @toomuchdesign 

I've did some research and came up with following findings running next server:

```js
DOMAIN_FROM_RUNTIME:
NEXT_PUBLIC_DOMAIN_FROM_RUNTIME: NEXT_PUBLIC_DOMAIN_FROM_RUNTIME
DOMAIN_FROM_CONFIG: DOMAIN_FROM_CONFIG
NEXT_PUBLIC_DOMAIN_FROM_CONFIG: NEXT_PUBLIC_DOMAIN_FROM_CONFIG
DOMAIN_FROM_DOTFILE:
NEXT_PUBLIC_DOMAIN_FROM_DOTFILE: NEXT_PUBLIC_DOMAIN_FROM_DOTFILE
```

Based on this it seems that `NEXT_PUBLIC_` filtering happens for dotenv vars + runtime vars, but not for vars coming from config.

toomuchdesign (approved):

Issue #154, opened 2021-02-18
next/head tags are not populated on server or client render


jesstelford:
## 🐛 Bug Report

Using Next.js' `next/head` component fails to inject tags into the `<head>` of a document when rendered within next-page-tester.

My motivation here is to test that meta tags and other SEO-related items (LDJSON, microformat, etc), are being rendered to the `<head>` correctly, but to do so I need to be able to see what's in the head.

## To Reproduce

Try rendering anything using `next/head`:

```javascript
import React from "react";
import Head from "next/head";

const Home = () => {
  return (
    <Fragment>
      <Head>
        <title>Server Side Title</title>
      </Head>
      <h1>Hello world</h1>
    </Fragment>
  );
};

export default Home;
```

Then check the output within a test:

```javascript
const { serverRender } = await getPage({
  route: "/",
});

serverRender();

const serverHtml = document.documentElement.outerHTML;
// <html data-reactroot=\\"\\"><head></head><body><div id=\\"__next\\"><h1>Hello world</h1></div></body></html>
```

Note while a `<head>` tag does exist, the expected `<title>Server Side Title</title>` does not.

## Expected behavior

`next-page-tester` should correctly populate the `<head>` with `next/head` rendered items.

## Link to repl or repo

Minimal reproduction: https://github.com/jesstelford/next-page-tester-bug-154

## Run `npx envinfo --preset jest`

Paste the results here:

```bash
  System:
    OS: macOS 10.15.7
    CPU: (4) x64 Intel(R) Core(TM) i5-7267U CPU @ 3.10GHz
  Binaries:
    Node: 12.20.1 - ~/.nvm/versions/node/v12.20.1/bin/node
    Yarn: 1.19.1 - ~/.config/yarn/global/node_modules/.bin/yarn
    npm: 6.14.9 - ~/dev/test/node_modules/.bin/npm
  npmPackages:
    jest: ^26.6.3 => 26.6.3
```

jesstelford:
I tried to do some debugging, but didn't get far.

I expected `updateHead` to be called here: https://github.com/toomuchdesign/next-page-tester/blob/0db34fe92de00df5e6cc40918470dfe53bbff68d/src/_document/render.tsx#L95-L102, but it never seems to be executed (the wrapping `renderPage` definitely is).

[...]

PR #156, opened 2021-02-19
fix: Default document rendering

Meemaw:

Fix default document rendering.

Fixes https://github.com/toomuchdesign/next-page-tester/issues/154

When `useDocument=false` we didn't properly render the Next.JS document (we skipped it) which meant that any `SEO` elements defined in custom `_app` or in `page` were not rendered.

This is wrong behaviour that is not in line with `useApp` where this flag only controls if we should use custom `_app` component, but not changing the rendering flow.

## What is the new behaviour?

`useDocument=false` is now inline with the `useApp=false` behaviour and document rendering flow is always correct.

### DIFF
--- a/src/__tests__/get-page-return/get-page-return.test.tsx
+++ b/src/__tests__/get-page-return/get-page-return.test.tsx
@@ -6,6 +6,7 @@ import path from 'path';
 import App from './__fixtures__/pages/_app';
 import Page from './__fixtures__/pages/page';
 import { expectDOMElementsToMatch } from '../__utils__';
+import { NEXT_ROOT_ELEMENT_ID } from '../../constants';
 
 // @NOTE These tests are not extensive.
 // They provide a rough idea of the values returned by getPage
@@ -20,15 +21,28 @@ describe('getPage() return', () => {
 
       const expectedPage = (
         <html>
-          <head />
+          <head>
+            <meta name="next-head-count" content="0" />
+            <noscript data-n-css=""></noscript>
+          </head>
           <body>
-            <div id="__next">
+            <div id={NEXT_ROOT_ELEMENT_ID}>
               <App Component={Page} />
             </div>
+            <script
+              id="__NEXT_DATA__"
+              type="application/json"
+              dangerouslySetInnerHTML={{
+                __html:
+                  '{"page":"/page","query":{},"buildId":"next-page-tester","props":{}}',
+              }}
+            ></script>
           </body>
         </html>
       );
-      const expectedHtml = ReactDOMServer.renderToString(expectedPage);
+
+      const expectedHtml = ReactDOMServer.renderToStaticMarkup(expectedPage);
+
       expect(html).toEqual(expectedHtml);
     });
   });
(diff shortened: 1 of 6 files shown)


src/_document/getDocumentFile.ts
```diff
@@ -8,15 +8,17 @@ export function getDocumentFile({
 }: {
   options: ExtendedOptions;
 }): NextDocumentFile {
-  const customDocumentFile = getPageFileIfExists<NextDocumentFile>({
-    options,
-    pagePath: DOCUMENT_PATH,
-  });
+  const { useDocument } = options;
```
Meemaw, line 11: This is exactly how `useApp` behaves.

toomuchdesign:
Hi @Meemaw, thanks for this PR and all the maintenance job during this days ❤️ .

I still haven't had the opportunity of fully digging into this pr. I'll do as soon as possible. Does it actually concern and fix #154?

At a first glance I'd expect this library to not render the `head` element when `useDocument === false` and consequently skipping and nested SEO element.

🙌

Meemaw:
> Hi @Meemaw, thanks for this PR and all the maintenance job during this days ❤️ .
> 
> I still haven't had the opportunity of fully digging into this pr. I'll do as soon as possible. Does it actually concern and fix #154?

Actually it seems that the issue is in the setup (perhaps missing `jest` patch) and does not require any changes on our side -- the reproducer works in our repo. 

> 
> At a first glance I'd expect this library to not render the `head` element when `useDocument === false` and consequently skipping and nested SEO element.

Hmm 🤔 I'm actually not sure which one I would prefer. On one hand I see many advantages to doing full Next.JS render even when `useDocument=false`:
- Simplified code flow internally / less branches to worry about
- Possibility to test `page`'s SEO elements even when `useDocument=false` e.g. `page` title takes precedence over `_document`'s title anyway
- More closely aligned with the main selling point of the library: `"The idea behind this library is to reproduce as closely as possible the way Next.js works without spinning up servers, and render the output in a local JSDOM environment."`

From what I see the main point of even preserving `useApp` and `useDocument` flags is possibility to avoid mocking complexity they might bring. E.g. in complex apps custom `_document`  might do 20 BE requests that would have to be mocked for every test.

How do you see it?

> 
> 🙌

toomuchdesign:
Hi @Meemaw,
I fully agree with your point of view about `useApp` and `useDocument` flags. I alse tend to see `useDocument` as a possible future escape hatch in case Next.js rewrites the way Document works and we were not able to make work it here.

Even without document, this library could be useful in quite a lot projects. Especially simple ones for which browser test would be overkill or too expensive.

src/__tests__/get-page-return/get-page-return.test.tsx
```diff
@@ -40,22 +54,31 @@ describe('getPage() return', () => {
...
               <App Component={Page} />
             </div>
+            <script
+              id="__NEXT_DATA__"
+              type="application/json"
+              dangerouslySetInnerHTML={{
+                __html:
+                  '{"page":"/page","query":{},"buildId":"next-page-tester","props":{}}',
+              }}
+            ></script>
```
toomuchdesign, line 78: Should we consider this output (along with `next-head-count` meta, `data-n-css` noscript..) implementation details not depending on our code and keep them out from out expectations?

If Next.js added a new next-tag in a minor version, our test would get red, wouldn't them?

toomuchdesign (commented):
If we went this direction we might consider taking an alternative testing strategy since I feel like we're exposing a lot of Next.js' implementation details as test expectation.

I currently don't have the full picture of our document tests, may we now have a bit of redundancy?

BTW, I like the way this approach simplify our code. Probably the cost of it gets reflected onto tests.

If you agree I'd keep the decision for after the upcoming release. We're currently tackling quite a few issues which I'd like to see fixed before proceeding further. In the meanwhile let's keep talking. 👍

Issue #153, opened 2021-02-17
Catch all route 


icyJoseph:
## 🐛 Bug Report

When there's a page `/[...all].tsx` trying to `getPage` on `/sub/[route].tsx` has unwanted results, because `[...all].tsx` is selected on this heuristic, from getPageObject/getRouteInfo.js

```
  const matchingRouteInfo = pagePaths
        .map((pagePath, index) => {
        const result = pathname.match(pagePathRegexes[index]);
        if (result) {
            return makeRouteInfo_1.makeRouteInfo({
                route,
                pagePath,
                routeRegexCaptureGroups: result.groups,
            });
        }
    })
        .filter((result) => Boolean(result))
        .sort((a, b) => a.paramsNumber - b.paramsNumber);
    // Return the result with less page params
    return matchingRouteInfo[0];
```

Because `paramsNumber` for both routes equals 1. The ordering I believe comes from `fastGlob`, which returns `[.` above all else,  see `getPageObject/getPagePaths.js`, a quick check through their package documentation doesn't seem to offer a way to rank paths by specificity.

## To Reproduce

- On a next js project, create `pages/[...all].tsx` and `pages/sub/[route].tsx`. 

- Try to `renderPage` with route on `/sub/route0`.

- `[...all].tsx` is rendered.

## Expected behavior

Catch all routes should have less preference.

osadi:
I've created a small reproduction repo here: https://github.com/osadi/next-page-tester

PR #157, opened 2021-02-19
fix: Use `getSortedRoutes` from next/next-server

icyJoseph:

Fixes https://github.com/toomuchdesign/next-page-tester/issues/153

Use `next` own route sorting solution. Found [here](https://github.com/vercel/next.js/blob/f06c58911515d980e25c33874c5f18ade5ac99df/packages/next/next-server/lib/router/utils/sorted-routes.ts).

This PR simply imports the sorting function, `getSortedRoutes`, and wraps the return value of `getPagePaths` with it.

The library uses its own sorting algorithm.

## What is the new behaviour?

Use `next` routing solution. What you get with `next`, you get with this library.

No.

Nada.

## Other information:

Some tests used this pattern `some/route/[slug]/index.js`, which is not accepted by `next`. I've updated these to be `some/route/[slug].js`

- [ x ] Added tests for a situation where a top level catch all is present. If a route matches a nested catch all, then that route should render, otherwise the top level catch all is used.


Meemaw:
Thanks for this @icyJoseph 👍 

Q: Was there a problem with adding a simple test fixture for the case you've also described in your issue? We can ignore custom `_document`, `_app`,... but I guess we would still like to have that covered.

icyJoseph:
> Thanks for this @icyJoseph 
> 
> Q: Was there a problem with adding a simple test fixture for the case you've also described in your issue? We can ignore custom `_document`, `_app` edge cases, and just focus the test on making sure it takes the correct page.

Not really. I just wanted to confirm that we could ignore `_document`, `_app` and `404` for instance, and just focus on this particular case.

I'll get on it!

icyJoseph:
I have implemented a couple of tests. I wonder if some documentation needs to be updated?

Meemaw (approved):
> I have implemented a couple of tests. I wonder if some documentation needs to be updated?

I don't think so. LGTM, but lets wait for @toomuchdesign thoughts.

toomuchdesign:
Hi @icyJoseph, hi @Meemaw,
thanks for the blazing job here! I'm totally up for reusing `getSortedRoutes` from Next.js

I have 2 points I'd like to hear your feedback about

### Refactoring route matching

I understand that `getSortedRoutes` sorts routes matching order even before knowing which route we're actually using. If so I guess we might refactor a few lines in [`getRouteInfo`](https://github.com/toomuchdesign/next-page-tester/blob/master/src/getPageObject/getRouteInfo.ts). Would  it make sense?

Instead of processing every existing route, generating a `routeInfo` object for every matching route, filtering and sorting, we might loop over `pagePath` and return with a `routeInfo` object as soon as we have a matching path. (since I understand `getSortedRoutes` should ensure that the first matching path is the route to match, right).

If this makes sense, we might also get rid of `paramsNumber` prop evaluation and type definition.

### Tests

I see that `routing-top-catch-all` tests replicate quite a lot of test cases already declared in existing `routing` tests (in their own file pointing to their). I'm definitely missing a lot of context, but I was wondering whether we could be able to shrink down these tests to the minimum necessary. The amount of tests supporting this library is quite extensive, and I think we should thrive to keep them down to a manageable number.

Shall we move `routing-top-catch-all` tests into `routing` tests folder? Since they need their own `pages` folder is it maybe better not to move them?

icyJoseph:
> Hi @icyJoseph, hi @Meemaw,
> thanks for the blazing job here! I'm totally up for reusing `getSortedRoutes` from Next.js
> 
> I have 2 points I'd like to hear your feedback about
> 
> ### Refactoring route matching
> I understand that `getSortedRoutes` sorts routes matching order even before knowing which route we're actually using. If so I guess we might refactor a few lines in [`getRouteInfo`](https://github.com/toomuchdesign/next-page-tester/blob/master/src/getPageObject/getRouteInfo.ts). Would it make sense?
> 
> Instead of processing every existing route, generating a `routeInfo` object for every matching route, filtering and sorting, we might loop over `pagePath` array and return with a `routeInfo` object as soon as we have a matching path. (since I understand `getSortedRoutes` should ensure that the first matching path is the right match, right?)
> 
> If this makes sense, we might also get rid of `paramsNumber` prop evaluation and type definition.

This is actually a good take. I was thinking of introducing `getSortedRoutes` with minimal impact on the code base, but I reckon that `getSortedRoutes` renders the `paramsNumber` approach unnecessary. I could take this one.

> ### Tests
> I see that `routing-top-catch-all` tests replicate quite a lot of test cases already declared in existing `routing` tests (in their own file pointing to their). I'm definitely missing a lot of context, but I was wondering whether we could be able to shrink down these tests to the minimum necessary. The amount of tests supporting this library is quite extensive, and I think we should thrive to keep them down to a manageable number.
> 
> Shall we move `routing-top-catch-all` tests into `routing` tests folder? Since they need their own `pages` folder is it maybe better not to move them?

I would argue that you already have good enough tests as it is. 

The root issue was a situation where a top catch all is chosen over a sub catch all. This is rather an edge case, which ends up catching other routes as well. I added the tests on a separate directory to avoid polluting the rest of tests. 

I will go through it once again, and try to narrow down the edge case furthermore, to basically two entries, one where the sub catch all is rendered, and one where the catch all kicks in.

toomuchdesign:
Thanks @icyJoseph! Feel free to ping me if you need any support/advice 👍

toomuchdesign (approved):
Seems excellent to me. Thanks so much @icyJoseph!

Issue #160, opened 2021-02-21
Using next/router throws "You should only use "next/router" inside the client side of your app"


Meemaw:
## 🐛 Bug Report

Using `SingletonRouter` exported from `next/router` does not work.

## To Reproduce

```
Import Router from 'next/router'

<button onClick={() => Router.replace('/')}
```

## Expected behavior

Should be possible to use `SingletonRouter` exported from `next/router`

### Note

This happens because we never call https://github.com/vercel/next.js/blob/canary/packages/next/client/router.ts#L138 which would initialise singleton router. However, just calling the function wont solve the issue. We need to somehow re-wire the initialisation there to use our "mocked" router instead of the default one.

I've been fighting `jest` a bit trying to achieve that but didn't succeed so far.

PR #161, opened 2021-02-21
fix: client side SingletonRouter

Meemaw:

Enables client side `SingletonRouter`  routing exported from `next/router`.

Fixes https://github.com/toomuchdesign/next-page-tester/issues/160

It is not possible to use client side `SingletonRouter`  routing exported from `next/router`.

No

### DIFF
--- a/src/__tests__/client-navigation/__fixtures__/PageA.tsx
+++ b/src/__tests__/client-navigation/__fixtures__/PageA.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import Link from 'next/link';
 import { UrlObject } from 'url';
-import { useRouter } from 'next/router';
+import Router, { useRouter } from 'next/router';
 
 type Props = {
   href: string;
@@ -10,14 +10,6 @@ type Props = {
 
 export const PageA = ({ href, hrefObject }: Props) => {
   const { replace } = useRouter();
-
-  const goToPageBstring = () => {
-    replace(href);
-  };
-
-  const goToPageBObject = () => {
-    replace(hrefObject);
-  };
 
   return (
     <div>
@@ -30,8 +22,19 @@ export const PageA = ({ href, hrefObject }: Props) => {
         <a>Go to B with Link (with object)</a>
       </Link>
 
-      <a onClick={goToPageBstring}>Go to B programmatically (with string)</a>
-      <a onClick={goToPageBObject}>Go to B programmatically (with object)</a>
+      <a onClick={() => Router.replace(href)}>
+        Go to B programmatically (next/router - with string)
+      </a>
+      <a onClick={() => Router.replace(hrefObject)}>
+        Go to B programmatically (next/router - with object)
+      </a>
+
+      <a onClick={() => replace(href)}>
+        Go to B programmatically (useRouter - with string)
+      </a>
+      <a onClick={() => replace(hrefObject)}>
+        Go to B programmatically (useRouter - with object)
+      </a>
     </div>
   );
 };
--- a/src/__tests__/client-navigation/client-navigation.test.tsx
+++ b/src/__tests__/client-navigation/client-navigation.test.tsx
@@ -10,11 +10,13 @@ const nextRoot = __dirname + '/__fixtures__';
 
 describe('Client side navigation', () => {
   describe.each`
-    title                                   | linkText
-    ${'using Link component (with string)'} | ${'Go to B with Link (with string)'}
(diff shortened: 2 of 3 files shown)


Meemaw:
This is a super draft implementation that demonstrated one possible way to do it. As we have to "re-wire" a locally scoped variable in a module I don't see a better way than to somehow "hack" ourselves into the initialisation.

Open question:
- Where should we call this logic? Is the current place okay?
- How to handle non-jest environments? Do we even have to worry about them? I'm a bit sceptical its possible to run the library in other test runners anyway.

toomuchdesign:
I know it might sound a bit punk but... how about not providing support `SingletonRouter`?  Is there anything `SingletonRouter` can do which `useRouter` can't?

Meemaw:
> I know it might sound a bit punk but... how about not providing support `SingletonRouter`? Is there anything `SingletonRouter` can do which `useRouter` can't?

No, but it means that projects that are using it have no change to use `next-page-tester`. I'm not saying its good usage (I'm against it personally) and it doesn't feel React'ish.

In a project I'm currently trying to integrate `next-page-tester` we use it all over the place and it would be impossible to get rid of it ATM 😢

toomuchdesign:
Hi @Meemaw,
sorry for keeping you waiting, I'haven't had much time to spend in extra screen activity lately.

If I get it right, here we're patching the default `next/router` entry point with an object created in `src/router/makeRouterMock.ts` by `makeRouterMock` function.

Since `makeRouterMock` replaces the mock instance (`SingletonRouter`) at every run, is the newly created mock always available through `next/router` default entry point? Or the mock is then one available at `jest.mock` time?

### Alternative 1
This PR seems quite all about Jest configuration. Shall we consider providing the minimum necessary API to let users mock `next/router` on their side?

### Alternative 2
Would it be much of a hassle wrapping the pages consuming `SingletonRouter` with a HOC providing the hook version of Next.js router?

Meemaw:
> ### Alternative 1
> This PR seems quite all about Jest configuration. Shall we consider providing the minimum necessary API to let users mock `next/router` on their side?
> 

I was thinking about that -- and tried it but... I couldn't get it to wrong. Actually any change to the current approach breaks it 😢 

> ### Alternative 2
> Would it be much of a hassle wrapping the pages consuming `SingletonRouter` with a HOC providing the hook version of Next.js router?

1. Yes, not really possible as the `SingletonRouter` is used across router helper utility functions that are spread all over the place ... 
2. I mentioned this in one of the PRs some time ago: we shouldn't require users to change their code in order to adopt/use to library. They will more likely just give up and not use it.

toomuchdesign:
Another punk idea. What if we tried to misuse `router` option to notify `jest` (or any other test runner) that `next/router` mock should be updated? After all we call it every time `SingletonRouter` should be updated.

Meemaw:
> Another punk idea. What if we tried to misuse `router` option to notify `jest` (or any other test runner) that `next/router` mock should be updated? After all we call it every time `SingletonRouter` should be updated.

Not sure what you mean by this 🤔

toomuchdesign:
I'll try to put down something :)

toomuchdesign:
Howdy @Meemaw?
I set up a demo branch to mock singletonRouter in user land by misusing `router` option:
https://github.com/toomuchdesign/next-page-tester/tree/fix-client-singleton-router-II

I'm sure it's improvable and I'd be happy if you validate it and take care of it. If you found it decent enough, we might merge new tests and provide this setup as docs in our `/example` folder.

PS. I'd like to keep commit `638f6bc`, since it helped me a lot understanding what tests were doing :)

Let me know! 👊

Meemaw:
Hey @toomuchdesign 

That looks promising 🎉 

Do you need to override each of the properties on `SingletonRouter` manually? If `SingletonRouter = router` doesn't work I think we can do this:


```javascript
Object.assign(SingletonRouter, router);
```

toomuchdesign:
I'd like to be wrong, but I'd say yes, you need to mutate `SingletonRouter` object.

toomuchdesign:
I remember `Object.assign(SingletonRouter, router)` causing issues with the way `SingletonRouter` is generated (some getter fun involved).

Meemaw:
> I remember `Object.assign(SingletonRouter, router)` causing issues with the way `SingletonRouter` is generated (some getter fun involved).

Managed to get it to work in https://github.com/toomuchdesign/next-page-tester/pull/173

Issue #147, opened 2021-02-12
Test fails because "Invalid hook call."


massimeddu-sonic:
## 💬 Questions and Help
Hello,

I'm facing an issue that I'm not able to solve with the following test:

```typescript
import { screen } from '@testing-library/react';
import { getPage } from 'next-page-tester';

describe('Home page', () => {
  it('renders without crashing', async () => {
    const { render } = await getPage({
      route: '/',
      dotenvFile: '.env.local'
    });
    render();
    expect(screen.getByRole('heading', { name: 'HomePage' })).toBeInTheDocument();
  });
```

I got this error

```
Home page › renders without crashing

    Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.

      22 |
      23 | export default function Popup(): JSX.Element {
    > 24 |   const [prefix, setPrefix] = useState<string>('');

```

with this component:

```typescript
[...]

export default function Popup(): JSX.Element {
  const [prefix, setPrefix] = useState<string>('');
[...]
```

Do you know I can solve this?

Thank you,

toomuchdesign:
Hi @massimeddu-sonic,
would you make sure that your issue is not the same as #118?

massimeddu-sonic:
Hi Andrea,

it looks like a different issue. I had that #118 issue and I've solved it using the patched version of jest, but then this issue arise.

Meemaw:
Are you using some context providers from third party libraries?

massimeddu-sonic:
Yes,

this is how it looks my `_app.tsx`

```typescript
import React from 'react';

import type { AppProps } from 'next/app';

import 'lib/styles/global.scss';
import { SWRConfig } from 'swr';

import CookiesPopup from '~/components/cookiespopup';
import { DownloadAppContextProvider } from '~/contexts/download-app.context';
import fetch from '~/lib/fetch';

const webApp = ({ Component, pageProps }: AppProps): JSX.Element => {
  return (
    <>
      <SWRConfig
        value={{
          fetcher: fetch,
        }}
      >
        <DownloadAppContextProvider>
          <Component {...pageProps} />
        </DownloadAppContextProvider>
        <CookiesPopup />
[...]

PR #162, opened 2021-02-21
fix: executeWithFreshModules when comsumed as package

toomuchdesign:

Bug fix

#147 

## What is the new behaviour?

Configure Jests's non-isolated modules from `testHelpers` inside a `beforeAll` callback.

No. Unless we consider users not using `initTestHelpers`.

## Other information:

Since the issue shows up only when `next-page-tester` is consumed as package, I haven't found a way to test this fix. Any advice is hugely appreciated. Also advices about monorepos :)

I though of keeping the non-isolated loop inside `utils` file, but I ended up thinking its place is `testHelpers`: after all Jest mocks should be setup before tests run. Any preference?

Most of this PR consist of replicating Emotion setup provided in #147. If we don't find a way to test this fix we might consider removing the newly added test.


Meemaw:
What is the actual fix here? Is it just moving the nonIsolatedModules to be setup earlier in the flow?

toomuchdesign:
Yep. That's all. I still haven't figured out what the issue was and I couldn't even cover it with a test.

fracture91:
I applied the meat of this PR (the `src/utils.ts` and `src/testHelpers.ts` changes) as a patch to my local node_modules and can confirm that it fixes the "invalid hook call" issue for me :+1: No new problems. https://github.com/toomuchdesign/next-page-tester/issues/147#issuecomment-779551551

toomuchdesign:
Screen Shot 2021-02-21 at 9 59 52 PM

I've never :)

toomuchdesign:
I added a `test:published` script to run tests against the published `next-page-tester` package itself. `npm` cli seemed to support us, after initial skepticism.

We cannot automate `test:published` execution in CI because tests are supposed to reflect a code state which hasn't still been published. BUT we might at least execute it manually when necessary.

src/testHelpers.ts
```diff
@@ -26,6 +27,17 @@ export function initTestHelpers() {
   if (typeof document !== 'undefined' && typeof afterEach === 'function') {
     afterEach(cleanup);
   }
+
+  if (typeof jest !== 'undefined') {
```
Meemaw, line 31: By moving this into `testHelpers` we are making usage of `initTestHelpers` mandatory for the `getPage` to work correctly. This might be a bit annoying and I'm sure some people might forget to do it (or intentionally avoid it) and open issues. What do you think?

Just curious: what happens if we move this into somewhere early in `getPage` -- does it work?

toomuchdesign: This was exactly my concern.
For the sake of our emotion setup, this would also work if we kept non-isolaated modules Jest loop in `utils` (executed at module loading time). I can't guarantee this would work in any context, though.

Since we're Jest mocking things, I though the best place was in a `beforeAll` callback. But `testHelpers` setup becomes mandatory, then.

I start to suspect we stumbled upon another Jest bug.

Meemaw: So yeah it's a decision we need to make. If we move it `initTestHelpers` I think we need to improve the DX as well by auto initing `testHelpers`.

I think its very off-putting when you want to try integrating a new library into a complex project and you see some errors. I think it should just work out of the box.

For advanced users that want to opt-out from this behaviour and control it themselves we can provide exactly the same approach as `@testing-library/react`.

toomuchdesign: > I think we need to improve the DX as well by auto initing testHelpers.

 You mean avoid letting users to drop these lines in their `setupFilesAfterEnv` files?
 
 ```
import { initTestHelpers } from 'next-page-tester';
initTestHelpers();
 ```

In order to make a step further we should hook into Jest and programmatically configure `setupFilesAfterEnv`. I've never tried. If possible, non-Jest (phantomatic?) users would still have to do it manually.

Or you're more thinking about detecting that proper env configuration is missing and alert users?

Meemaw: Yes they would be able to omit those lines, but more importantly -- new users wont see errors in case they forget it. Ofcourse they would go to the repo and in the `README.md` read they need to use it, but we can them (and ourselves -- by receiving less issues) a favour if we wire everything up ourselves.

So all that is needed is something like this:
- Rename current `index.ts` -> `pure.ts`
- Add a new `index.ts` with content similar to this


```ts
import { initTestHelpers } from 'next-page-tester'
export * from './pure'

if(!process.env.NEXT_PAGE_TESTER_SKIP_TEST_INIT) {
    initTestHelpers();
}
```

Meemaw (approved):

src/testHelpers.ts
toomuchdesign: I see now. I wasn't aware of such a trick in testing-library. Would it somehow also work for `setupFilesAfterEnv`? I'm totally open to such a solution if we found a proper way to test it.

Off topic: I'm a bit concerned about how we now replace/filter runtime (runtime only) env vars. We get closer to what's expected in a Next.js runtime routine but when in client environment (which we consider the base one) we're practically filtering everything out. Let's see if it break things :)

Meemaw: Not sure what you mean with `setupFilesAfterEnv`. This is just jest's specific hook to initialise some stuff. 

Whats great about the trick I've proposed and testing-library is using is that it works by default with every testing framework that supports `afterEach` callback (most of them do).

No need to do anything with `setupFilesAfterEnv`.

When user imports `getPage`:

```js
import { getPage } from 'next-page-tester'
```

Module will get executed and setup the `afterEach` callbacks and mocks.

toomuchdesign: Oki, let's give it a try. I'll merge this PR and maybe release v0.21.0 so that we can better focus on what's next. The great thing about living in v0 😄

PR #172, opened 2021-02-24
fix: revive nonIsolatedModules

Meemaw:

Re-introduce `nonIsolatedModules` option. This enables module mocking in "server" context.

It is not possible to mock a module (e.g. SDK) in "server" context. Because test is executed in "client" context any mocks configured in test code are not applied to "server" context.

## What is the new behaviour?

Add ability to mark such modules as "nonIsolated". Identity of those modules is then preserved in both contexts and mocking is possible.

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ React element of the application.
 | **nextRoot**                   | Absolute path to Next.js root folder                                               | `string`                          | _auto detected_ |
 | **dotenvFile**                 | Relative path to a `.env` file holding [environment variables][next-docs-env-vars] | `string`                          | -               |
 | **wrapper**                    | Map of render functions. Useful to decorate component tree with mocked providers.  | `{ Page?: NextPage => NextPage }` | -               |
+| **nonIsolatedModules**         | List of modules that should preserve identity between client and server context.   | `string[]`                        | []              |
 
 ## Skipping Auto Cleanup & Helpers Initialisation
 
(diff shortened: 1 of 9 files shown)


Meemaw:
Tried to upgrade from `0.19.0` -> `0.21.0` in my project but realised all my mocks started working 😂

README.md
```diff
@@ -158,6 +159,25 @@ Under [examples folder][examples-folder] we're documenting the testing cases whi
 
 ## FAQ
 
+### How do I mock API calls in my data fetching methods?
+
+Because `next-page-tester` isolates modules between "client" and "server" context mocks that are created in test (client context) wont execute in data fetching methods (server context).
```
toomuchdesign, line 164:
```suggestion
Since `next-page-tester` isolates modules between "client" and "server" context, mocks created in test (client context) won't execute in data fetching methods (server context).
```

README.md
```diff
@@ -106,6 +106,7 @@ React element of the application.
 | **nextRoot**                   | Absolute path to Next.js root folder                                               | `string`                          | _auto detected_ |
 | **dotenvFile**                 | Relative path to a `.env` file holding [environment variables][next-docs-env-vars] | `string`                          | -               |
 | **wrapper**                    | Map of render functions. Useful to decorate component tree with mocked providers.  | `{ Page?: NextPage => NextPage }` | -               |
+| **nonIsolatedModules**         | List of modules that should preserve identity between client and server context.   | `string[]`                        | []              |
```
toomuchdesign, line 109: What if we removed the negation? Something like `preserveModuleIdentity`?

Same could apply for our internal methods like: `unIsolatePredefinedJestModules ` , `unIsolateJestModules `..

toomuchdesign (commented):
Hi @Meemaw, thanks for documenting this. I left just 2 comments, but I think it's good to go.

README.md
Meemaw: Hmm. Yeah naming this is confusing.... What about something less abstract for users to understand: `sharedModules` -- since their identity is "shared" between `client` and `server`.

toomuchdesign: Ok. Let's go with `sharedModules`!

Meemaw: Done 👍

toomuchdesign (approved):

PR #174, opened 2021-02-26
feat: Polyfill, fetch on server-env

icyJoseph:

Hi,

I encountered an issue with `fetch` on the server side. Formally Next.js does a polyfill for fetch on the server environment, using
node-fetch.

This PR aims to void introducing coupling to the `next/next-server` package, which would be caused by importing 'next/dist/next-server/server/node-polyfill-fetch'.

Additionally, importing the `node-polyfill-fetch` obscures the clean up after simulating rendering on server side.

- Since `cross-fetch` is already installed in the project it can be used to polyfill.

- The polyfill with `cross-fetch` is only done if `global.fetch` is not defined, to avoid overriding user land definitions for `fetch`.

In this `next` page:

```js
import Head from "next/head";
import styles from "../styles/Home.module.css";

export default function Home({ pokemon }) {
  return (
    <div>
      <Head>
        <title>{pokemon.name}</title>
      </Head>

      <main className={styles.main}>
        <h1 className={styles.title}>{pokemon.name}</h1>
      </main>

      <footer>
        <img src={pokemon.sprites.front_default} alt="a11y" />
      </footer>
    </div>
  );
}

export async function getServerSideProps() {
  const res = await fetch("https://pokeapi.co/api/v2/pokemon/1");
  const pokemon = await res.json();
  if (pokemon) {
    return { props: { pokemon } };
  } else {
    return { props: { pokemon: null } };
  }
}

```

Cannot be tested with `next-page-tester`, without stubbing or mocking `jest` on the node environment.

A simple test such as:

```js
import { screen } from "@testing-library/react";
import { getPage } from "next-page-tester";

describe("App", () => {
  it("renders without crashing", async () => {
    const { render } = await getPage({
      route: "/"
    });

    render();
    expect(screen.getByText("bulbasaur")).toBeInTheDocument();
  });
});

```

Won't be able to find `fetch` though it is only used on `getServerSideProps`. For better or worse, `next` polyfills `fetch` server side, and `next-page-tester` shouldn't disrupt that.

```bash
 FAIL  tests/index.test.js
  App
    ✕ renders without crashing (84 ms)

  ● App › renders without crashing

    ReferenceError: fetch is not defined

      27 |
      28 | export async function getServerSideProps() {
    > 29 |   const res = await fetch("https://pokeapi.co/api/v2/pokemon/1");
         |               ^
      30 |   const pokemon = await res.json();
      31 |   if (pokemon) {
      32 |     return { props: { pokemon } };

      at getServerSideProps (pages/index.js:29:15)
      at node_modules/next-page-tester/dist/fetchData/fetchPageData.js:93:71
      at Object.executeAsIfOnServer (node_modules/next-page-tester/dist/server.js:29:22)
      at Object.fetchPageData [as default] (node_modules/next-page-tester/dist/fetchData/fetchPageData.js:93:45)
      at Object.fetchRouteData (node_modules/next-page-tester/dist/fetchData/fetchRouteData.js:13:51)
      at Object.getPageInfo (node_modules/next-page-tester/dist/page/getPageInfo.js:20:22)
      at getPage (node_modules/next-page-tester/dist/getPage.js:64:38)
      at Object.<anonymous> (tests/index.test.js:6:24)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 total
Snapshots:   0 total
Time:        1.215 s, estimated 2 s
```

## What is the new behaviour?

As explained above this PR attempts to polyfill on server side, only if `fetch` is not present.

```js
import fetch from 'cross-fetch';

// other imports

function hideBrowserEnv(): () => void {
  const tmpWindow = global.window;
  const tmpDocument = global.document;
  const tmpFetch = global.fetch; // <---
  global.fetch = tmpFetch || fetch; // <---

  // @ts-expect-error its okay
  delete global.window;
  // @ts-expect-error its okay
  delete global.document;
  setNextRuntimeConfig({ runtimeEnv: RuntimeEnvironment.SERVER });
  setEnvVars({ runtimeEnv: RuntimeEnvironment.SERVER });

  return () => {
    global.window = tmpWindow;
    global.document = tmpDocument;
    global.fetch = tmpfetch; <-- clean up

    setNextRuntimeConfig({ runtimeEnv: RuntimeEnvironment.CLIENT });
    setEnvVars({ runtimeEnv: RuntimeEnvironment.CLIENT });
  };
}
```
...

No, because we'd use `cross-fetch` as polyfill only if the user has not specified any.

Nada.

## Other information:

As [polyfilled by `next`](https://github.com/vercel/next.js/blob/canary/packages/next/next-server/server/node-polyfill-fetch.js)

- [ x ] Tests for the changes have been added

### DIFF
--- a/src/__tests__/__utils__/fetcher.ts
+++ b/src/__tests__/__utils__/fetcher.ts
@@ -1,0 +1,2 @@
+export const fetcher = (resource: string) =>
+  fetch(resource).catch(() => resource);
--- a/src/__tests__/__utils__/index.ts
+++ b/src/__tests__/__utils__/index.ts
@@ -10,3 +10,4 @@ export {
 export { expectToBeDefault404Page } from './expectToBeDefault404Page';
 export { renderWithinNextRoot } from './renderWithinNextRoot';
 export { silenceConsoleError } from './silenceConsoleError';
+export { fetcher } from './fetcher';
--- a/src/__tests__/page-data-fetching/__fixtures__/pages/ssr/[id].js
+++ b/src/__tests__/page-data-fetching/__fixtures__/pages/ssr/[id].js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { PropsPrinter, sleep } from '../../../../__utils__';
+import { fetcher, PropsPrinter, sleep } from '../../../../__utils__';
 
 export default function ssr_$id$(props) {
   return (
@@ -12,6 +12,7 @@ export default function ssr_$id$(props) {
 
 export async function getServerSideProps(ctx) {
   await sleep(1);
+  await fetcher('resource');
   return {
     props: ctx,
   };
--- a/src/server.ts
+++ b/src/server.ts
@@ -7,8 +7,8 @@ import { setEnvVars } from './setEnvVars';
 function hideBrowserEnv(): () => void {
   const tmpWindow = global.window;
   const tmpDocument = global.document;
-  const tmpfetch = global.fetch;
-  global.fetch = fetch;
+  const tmpFetch = global.fetch;
+  global.fetch = tmpFetch || fetch;
 
   // @ts-expect-error its okay
   delete global.window;
@@ -20,7 +20,7 @@ function hideBrowserEnv(): () => void {
   return () => {
     global.window = tmpWindow;
     global.document = tmpDocument;
-    global.fetch = tmpfetch;
+    global.fetch = tmpFetch;
 
     setNextRuntimeConfig({ runtimeEnv: RuntimeEnvironment.CLIENT });
     setEnvVars({ runtimeEnv: RuntimeEnvironment.CLIENT });
--- a/src/server.ts
+++ b/src/server.ts
@@ -1,3 +1,5 @@
+import fetch from 'cross-fetch';
+
 import { RuntimeEnvironment } from './constants';
 import setNextRuntimeConfig from './setNextRuntimeConfig';
 import { setEnvVars } from './setEnvVars';
@@ -5,6 +7,8 @@ import { setEnvVars } from './setEnvVars';
 function hideBrowserEnv(): () => void {
   const tmpWindow = global.window;
   const tmpDocument = global.document;
+  const tmpfetch = global.fetch;
+  global.fetch = fetch;
 
   // @ts-expect-error its okay
   delete global.window;
@@ -16,6 +20,8 @@ function hideBrowserEnv(): () => void {
   return () => {
     global.window = tmpWindow;
     global.document = tmpDocument;
+    global.fetch = tmpfetch;
+
     setNextRuntimeConfig({ runtimeEnv: RuntimeEnvironment.CLIENT });
     setEnvVars({ runtimeEnv: RuntimeEnvironment.CLIENT });
   };


toomuchdesign:
Hi @icyJoseph!
Thanks for filing the PR! I've never thought about `fetch` not being polyfilled by `next-page-tester`, but I'm also reasoning about the opportunity of polyfilling by default.

I agree we should most of the times aim to replicate the same environment provided by Next.js, but in our case it's about testing and I'd usually be inclined to avoid tests to perform real network requests.

In other words, from my point of view `ReferenceError: fetch is not defined` error is a quick alert signaling that something is not properly mocked. Of course there are further mocking strategies involving mocking at the network layer (like `msw`) and in this case user might want to have a fully functional `fetch` object. Therefore, yep, this is a debatable subject :)

@icyJoseph Do you happen to often perform actual network requests from within your Next.js tests?
@Meemaw what are your 2 cents?

Meemaw:
> @Meemaw what are your 2 cents?

- One one hand its cool that this would work out of the box
- On the other hand I always stub all network requests in there kind of tests so error being thrown might be useful

I think getting some feedback from users (when we have more of them 🙂 ) would be useful

icyJoseph:
I'd say nearly every developer who's going to write tests ends up mocking their network requests. 

While implementing this I had in mind only to not break the out of the box behavior given by Next, but you do make clear points.

Perhaps it would be better to print out a complementary message, telling users that `fetch` is not poly filled, and that way nudge them in the right direction. Or include this _GOTCHA_ in the documentation instead.

It may not help seasoned developers as much, but those learning could benefit a lot from clear logs. The JEST printout, `ReferenceError: fetch is not defined`, might throw beginners off.

toomuchdesign:
What if we started to add a couple lines of documentation? :)

toomuchdesign:
Hi @icyJoseph,
I've tried to fill a few documentation gaps about this error and mocking network requests. This was definitely a missing point, thanks for raising the issue. I'd be happy to have your feedback about it.

I also realised that `fetch` polyfilling could be a sort of grey area for our users. It should be necessary when mocking at network layer and probably useless when mocking at global `fetch` object level or directly at user land modules level. And I'm probably overlooking further mocking strategies :). What strategy use in your projects, @icyJoseph?

Couple of question about this PR.

Are getting `ReferenceError: fetch is not defined` error on  "server" `fetch` calls only or anywhere `fetch` is called in code? I don't remember whether JSDOM actually polyfills global `fetch` object, but I'd expect it to be equally available or not in both client/server environments.

If this is right, couldn't we lift `cross-fetch` up to `src/testHelpers.ts`? Or there is a specific reason why it's currently in `server.ts`?

I was also imagining a different approach consisting of providing a `fetch` polyfill consisting of no functionality but a simple user warning. But we should evaluate the implications of it.

BTW, thanks for raising attention on this point, let's find out the right step to take 👍

icyJoseph:
Hi, 

Sorry I was quite busy this week. 

> Hi @icyJoseph,
> I've tried to fill a few documentation gaps about this error and mocking network requests. This was definitely a missing point, thanks for raising the issue. I'd be happy to have your feedback about it.
> 

I think it looks good. 

> I also realised that `fetch` polyfilling could be a sort of grey area for our users. It should be necessary when mocking at network layer and probably useless when mocking at global `fetch` object level or directly at user land modules level. And I'm probably overlooking further mocking strategies :). What strategy use in your projects, @icyJoseph?
> 

We poly-fill `global.fetch` with `whatwg-fetch`, but to mock network request we use [msw](https://github.com/mswjs/msw).

I'll quote my colleague  @JohanGustafsson91 on why we use `whatwg-fetch`, maybe he can expand a bit more on our mocking strategy as well.

> Next seems to polyfill fetch with whatwg-fetch in the browser environment and node-fetch in the server/node environment. Unit tests are running in the node environment and the node-fetch polyfill does not support relative api routes. Therefore we need to polyfill with whatwg-fetch in the unit tests.

So I think, in our case, even the Next polyfill was against our desired behavior.

> Couple of question about this PR.
> 
> Are getting `ReferenceError: fetch is not defined` error on "server" `fetch` calls only or anywhere `fetch` is called in code? I don't remember whether JSDOM actually polyfills global `fetch` object, but I'd expect it to be equally available or not in both client/server environments.
> 

`ReferenceError: fetch is not defined` happens in both server and client side. 

> If this is right, couldn't we lift `cross-fetch` up to `src/testHelpers.ts`? Or there is a specific reason why it's currently in `server.ts`?
> 

I was trying to isolate this to the render on server code, as much as possible. The next-server polyfill had to be isolated like this, otherwise it'd inject `node-fetch` into the global object. When I saw that `cross-fetch` was used I must have been already locked in keeping changes to this file only.

Given that the error happens both client and server side, we could hoist this to `src/testHelpers.ts` and polyfil if `global.fetch` is not already defined.

> I was also imagining a different approach consisting of providing a `fetch` polyfill consisting of no functionality but a simple user warning. But we should evaluate the implications of it.
> 
> BTW, thanks for raising attention on this point, let's find out the right step to take

Right, because if not done correctly maybe their tests won't fail, or even if they do fail errors might be harder to trace. Some might miss the warning completely. 

All in all, I think having this properly documented is the best approach so far. 

Most developers do read through the `ReadME` before installing packages, and the rest will eventually do so when things go wrong. Poly-filling `node-fetch` with `cross-fetch` might lead in turn to its own unique set of problems and user complaints. 

Looking back, the selling point of this PR is that upon installing it and writing your first test, things just work. How much is that valued by end users is anyone's guess :) In our case, because we have [`msw`](https://github.com/mswjs/msw), we would have never even thought about what's happening to fetch.

Issue #175, opened 2021-02-26
next/head doesn't update DOM after initial render


toomuchdesign:
## 🐛 Bug Report

`next/head` doesn't update DOM after initial render (eg `useEffect ` callback).

## To Reproduce

Render a new `<Head>` element after mount with `useEffect`.

## Expected behavior

Newly rendered head element should be rendered in DOM.

## Link to repl or repo

- https://github.com/jesstelford/next-page-tester-bug-154
- #154

## Run `npx envinfo --preset jest`

```bash
  System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
  Binaries:
    Node: 14.15.4 - ~/.nvm/versions/node/v14.15.4/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v14.15.4/bin/yarn
    npm: 6.14.10 - ~/.nvm/versions/node/v14.15.4/bin/npm
  npmPackages:
    jest: ^26.6.3 => 26.6.3
```

PR #177, opened 2021-02-27
fix: next/head doesn't update DOM after initial render

toomuchdesign:

Bug fix

#175 

## What is the new behaviour?

Currently just adding a failing test.


toomuchdesign:
Anybody feel free to start from this pr to throw a fix for #177 :)

PR #178, opened 2021-02-28
fix(env): pass-through NODE_ENV to client

Meemaw:

Pass through `NODE_ENV` to the client

`NODE_ENV` is not available in `client` context which is:
- not in-line with `Next.JS`
- makes it hard to have conditional logic based on `process.env.NODE_ENV === 'test'`

## What is the new behaviour?

`NODE_ENV` is passed through to the client context

No


toomuchdesign (approved):
Good catch: we might need to extend what we pass through client render in future, and this PR keeps us covered 👍

PR #187, opened 2021-03-08
chore: bump next version to v10.0.8

toomuchdesign:

Chore

## What is the new behaviour?

Update Next dev and peer version to `^10.0.8` and update router mock to include new typed props

Yes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,10 +163,11 @@ Under [examples folder][examples-folder] we're documenting the testing cases whi
 
 `next-page-tester` focuses on supporting only the last major version of Next.js:
 
-| next-page-tester | next.js |
-| ---------------- | ------- |
-| v0.1.0 - v0.7.0  | v9.X.X  |
-| v0.8.0 +         | v10.X.X |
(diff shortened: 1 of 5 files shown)


toomuchdesign:
It seems I can't reproduce the CI failure on my machine. Need investigation.

toomuchdesign:
Update: it's a timeout error. We should make sure whether it's an issue introduced with async configuration loading, or we should just increase timeout threshold.

toomuchdesign:
I figured what's causing x4 execution time.

`loadConfig` makes a new [`loadWebpackHook` call](https://github.com/vercel/next.js/blob/33255b7f4ed6dd2608d6a6590d137353b6f19f67/packages/next/next-server/server/config.ts#L397) which seems to be the origin of the issue.

toomuchdesign:
With this hack makes me really sad, it's definitely an order of magnitude worst then any workaround we had to put up in this project :) @Meemaw I'll let to you the final word.

An alternative solution might consist of re-implementing config loading.

PR #212, opened 2021-05-10
feat: Add custom wrapper for _app

### LINKED ISSUE #210: Custom wrapper for _app
## 🚀 Feature Proposal

It would be great to have a way to customize _app. In source code I've seen that there is already a wrapper for `Page` and it have types for custom app.

## Motivation

In my project we use `next-translate` (via webpack plugin) and keep some layout  (header, footer) `in _app.tsx`. AFAIK I can't use webpack plugin to load locales during testing with `next-page-tester`, so I need to wrap app in provider and load them manually. I tried to wrap `Page` by using `wrapper.Page` but then my header and footer are left without locales.

I can try to implement this and submit a PR but I definitely would need help with it. I've already tried to modify this package but with no luck.

## Example

Something like this:
```typescript
const { render } = await getPage({
    route: "/",
    useDocument: true,
    wrapper: {
        App: App => {
             return appWithI18N(App);
        },
    },
});
```
or like this:
```typescript
const { render } = await getPage({
    route: "/",
    useDocument: true,
    wrapper: {
        App: App => {
             App.getInitialProps = async () => ({
                pageProps: {
                    ...(await loadNamespaces()),
                }
             });

             return App;
        },
    },
});
```

evocateur:
Closes #210

Adds a `wrapper.App` option, similar to the existing `wrapper.Page`.

Only a `Page` wrapper is available, which is certainly useful, but insufficient for certain behaviors.

## What is the new behaviour?

#210

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,18 +104,18 @@ React element of the application.
 
 ## Options
 
-| Property                       | Description                                                                        | type                              | Default         |
-| ------------------------------ | ---------------------------------------------------------------------------------- | --------------------------------- | --------------- |
-| **route** (mandatory)          | Next route (must start with `/`)                                                   | `string`                          | -               |
-| **req**                        | Enhance default mocked [request object][req-docs]                                  | `req => req`                      | -               |
-| **res**                        | Enhance default mocked [response object][res-docs]                                 | `res => res`                      | -               |
-| **router**                     | Enhance default mocked [Next router object][next-docs-router]                      | `router => router`                | -               |
-| **useApp**                     | Render [custom App component][next-docs-custom-app]                                | `boolean`                         | `true`          |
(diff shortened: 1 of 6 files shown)


toomuchdesign:
Hi @evocateur,
thanks for the PR! 🙌 

The CI seems to be failing because of a formatting issue. Would you mind running `npm run prettier:fix` and commit again the result?

Cheers!

evocateur:
@toomuchdesign Thanks, sorry about that! Trying to do too many things at once, forgot a fundamental step 😅

Meemaw (approved):
LGTM

toomuchdesign (approved):
Looks good to me! Thanks @evocateur! I'll leave it here for a little while to let anybody interested to check out 👍

toomuchdesign:
@all-contributors please add @evocateur for code

PR #241, opened 2021-07-25
fix: add next 11 as valid peer dependency

DoctorDerek:

While there shouldn't be any breaking changes or compatibility issues with Next 11, the package won't install without the --force flag.

Peer dependency is Next ^10

## What is the new behaviour?

Peer dependency is Next ^10 || Next ^11

No

- [ N/A ] Tests for the changes have been added
- [ N/A ] Docs have been added / updated


DoctorDerek:
Hey @toomuchdesign, thanks for merging the PR!

This might be a good time to push another release, since it will allow end users to start using the package with Next 11.

That's not mission critical for me at all, but I'm sure there are users out there 😀

Issue #245, opened 2021-08-12
Compatibility with next@11.1.0


dferber90:
I'm getting the following error with `next-page-tester@0.27.2` and `next@11.1.0`

```
 FAIL   xxx  tests/imprint.test.tsx
  ● Test suite failed to run

    Cannot find module 'next/dist/next-server/server/config' from '../../node_modules/next-page-tester/dist/nextConfig.js'

    Require stack:
      /Users/xxx/node_modules/next-page-tester/dist/nextConfig.js
      /Users/xxx/node_modules/next-page-tester/dist/utils.js
      /Users/xxx/node_modules/next-page-tester/dist/getPageObject/makeRouteInfo.js
      /Users/xxx/node_modules/next-page-tester/dist/getPageObject/index.js
      /Users/xxx/node_modules/next-page-tester/dist/page/getPageInfo.js
      /Users/xxx/node_modules/next-page-tester/dist/page/index.js
      /Users/xxx/node_modules/next-page-tester/dist/getPage.js
      /Users/xxx/node_modules/next-page-tester/dist/pure.js
      /Users/xxx/node_modules/next-page-tester/dist/index.js
      tests/getPages.ts
      tests/imprint.test.tsx

      at Resolver.resolveModule (../../node_modules/jest-resolve/build/resolver.js:311:11)
      at Object.<anonymous> (../../node_modules/next-page-tester/dist/nextConfig.js:7:34)
```

It seems like the `next/dist/next-server/server/config` file has changed to `next/dist/server/config` so the path needs to be updated:

I found these occurrences but there might be more:

https://github.com/toomuchdesign/next-page-tester/blob/f3ad5cd777e08790009a7c0820f07a65c26d5b3b/src/nextConfig.ts#L1

https://github.com/toomuchdesign/next-page-tester/blob/f3ad5cd777e08790009a7c0820f07a65c26d5b3b/src/nextConfig.ts#L15

toomuchdesign:
Next.js v11.1 support should available with v0.28. Thanks for raising the issue 👍

dferber90:
That's great to hear! Thanks for handling it so fast! 🙏

PR #246, opened 2021-08-12
chore: next-v11.1

Meemaw:

Fixes https://github.com/toomuchdesign/next-page-tester/issues/245

Cannot use `next-page-tester` with 11.1

## What is the new behaviour?

...


Divlo (approved):
Thanks for this PR! :+1: 

Overall LGTM, just the compatibility information for the current version that does not support `next@11.1.0`is unclear (see my suggestion).

src/__tests__/setup.ts
```diff
@@ -10,3 +10,5 @@ afterEach(() => {
     });
   }
 });
+
+jest.setTimeout(15000);
```
toomuchdesign, line 14: The timeout issue was introduced a few Next versions ago since Next.js bootstraps Webpack at configuration loading time. In order to prevent this we temporarily mocked out the module responsible for loading webpack in:  `src/nextConfig.ts`.

```
  const configUtils = require('next/dist/next-server/server/config-utils');
```
should become:
```
const configUtils = require('next/dist/server/config-utils');
```
I'd also ask if you could add a next-page-tester console warning in the catch clause to help detecting future internal changes.

Otherwise we might consider dropping Next.js config loading and implement our own without webpack. We might also double check to see whether configuration loading module now provides a way to skip webpack

src/utils.ts
```diff
@@ -210,10 +210,10 @@ export function setNextImageConfiguration(): void {
...
-    deviceSizes: config.images.deviceSizes,
-    imageSizes: config.images.imageSizes,
-    path: config.images.path,
-    loader: config.images.loader,
-    domains: config.images.domains,
+    deviceSizes: config.images?.deviceSizes,
+    imageSizes: config.images?.imageSizes,
+    path: config.images?.path,
+    loader: config.images?.loader,
+    domains: config.images?.domains,
```
toomuchdesign, line 217: Would it be feasible covering this optional chaining logic branch with a test to get back to 100% test coverage?

src/setEnvVars.ts
```diff
@@ -76,6 +76,8 @@ export function setEnvVars({
   const { env: envVarsFromConfig } = getNextConfig();
   // Runtime and dotfile env vars are scoped by environment (via NEXT_PUBLIC_ prefix),
   // while env vars coming from next.config.js are available in both environments
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore NODE_ENV expected, but we are passing it so its okay
```
toomuchdesign, line 80: if we really have to skip type checking we can do it with a `// @ts-expect-error {description}`. But it would be better if we could remove `src/next-env.d.ts`, if possible.

src/next-env.d.ts
```diff
@@ -0,0 +1,6 @@
+/// <reference types="next" />
+/// <reference types="next/types/global" />
+/// <reference types="next/image-types/global" />
+
+// NOTE: This file should not be edited
+// see https://nextjs.org/docs/basic-features/typescript for more information.
```
toomuchdesign, line 6: What is it for? tests seems to be green even without it. This also causes the type error in `src/setEnvVars.ts`.

toomuchdesign (changes requested):
Hi @Meemaw! Thanks for putting together this PR.
I left a few comments which I hope might help fixing a few runtime issues.

Feel free to ping me if you need support! (and remember you have push rights to this repo 😄 )

src/next-env.d.ts
Meemaw: Typescript fails without this.

toomuchdesign: I see now. It only fails when `tsc --noEmit`. No issues at tests or build time. So this is a new file Next.js users are supposed to keep in their userland projects, right? Using `skipLibCheck` means that all `next-page-tester` would be also forced to enable the option?

In the worst case we can keep it and `// @ts-expect-error` the error. How do you see it?

src/setEnvVars.ts
toomuchdesign: Or.. another strategy: couldn't we just declare `NODE_ENV` prop :)? (I know we already provide it implicitly)

```ts
  process.env = {
    NODE_ENV: process.env.NODE_ENV,
    ...getEnvVarsByEnvironment(baseEnvVars)[runtimeEnv],
    ...envVarsFromConfig,
  };
```

toomuchdesign: Or (2) we might make our `EnvVars` type more specific to extend `ProcessEnv`?

toomuchdesign:
This evening I might have a few time to bring forward this PR. In case I'll merge this PR into feature branch and continue from there.

toomuchdesign:
Hi @Meemaw :) I updated the branch according to what mentioned in comments. Let me know how do you see the pr.

toomuchdesign (approved):

Issue #263, opened 2021-09-22
Next 11.1.2 throws `TypeError: DocumentComponent.renderDocument is not a function`


fracture91:
## 🐛 Bug Report

Using `next-page-tester` 0.28.0 with `next` 11.1.2 (and possibly 11.1.1?) throws the following error with `useDocument: true`:

```
    TypeError: DocumentComponent.renderDocument is not a function

      4 | describe("404 page", () => {
      5 |   it("has a 404 status code", async () => {
    > 6 |     const { render } = await getPage({ route: "/asdf1234", useDocument: true })
        |                        ^
      7 |     render()
      8 |     expectStatusCode(404)
      9 |   })

      at node_modules/next-page-tester/dist/_document/render.js:96:34
      at Object.executeAsIfOnServer (node_modules/next-page-tester/dist/server.js:33:16)
      at getPage (node_modules/next-page-tester/dist/getPage.js:70:31)
      at Object.<anonymous> (__tests__/pages/404.js:6:24)
```

It looks like it's reaching this line of code https://github.com/toomuchdesign/next-page-tester/blob/bbde6b3714805c3e2d002f390ea28fa25aa6f61e/src/_document/render.tsx#L151

But the `renderDocument` function was deleted in this commit :(

https://github.com/vercel/next.js/commit/08a24780b01608e16b7733d31bda2aaf36fd1750#diff-5d810eb5cd68f6a6a3597e5472015f0c199e21d97cdabc9f7d5ac3fa85adc369L179

## Reproduction

I'm guessing it's trivial enough to bump your `next` version and see your tests fail

## Expected behavior

No errors :)

## Jest patch

Have you patched Jest as [described here](https://github.com/toomuchdesign/next-page-tester#optional-patch-jest)?

- [ ] yes
- [x] no, because I'm using Jest 27

## Run `npx envinfo --preset jest`

Paste the results here:

```bash
  System:
    OS: macOS 11.4
    CPU: (8) arm64 Apple M1
  Binaries:
    Node: 15.11.0 - ~/.nvm/versions/node/v15.11.0/bin/node
    Yarn: 1.22.5 - /opt/homebrew/Cellar/yvm/4.1.1/shim/yarn
    npm: 7.6.0 - ~/.nvm/versions/node/v15.11.0/bin/npm
  npmPackages:
    jest: ^27.0.0 => 27.0.4 
```

toomuchdesign:
[...]

PR #264, opened 2021-09-22
fix: _document rendering with next v11.1.1+

toomuchdesign:

Bug fix

#263 

## What is the new behaviour?

Let's see :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "next-page-tester",
-  "version": "0.28.1",
+  "version": "0.29.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "next-page-tester",
-      "version": "0.28.1",
+      "version": "0.29.0",
       "license": "ISC",
       "dependencies": {
         "@types/express": "^4.17.13",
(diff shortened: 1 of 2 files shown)


toomuchdesign:
This PR:

- updates Next.js to v11.1.2
- disable tests using `useDocument` option

The idea is to see whether we can ship a temporary `next-page-tester` version which doesn't support `useDocument` option while we find a better working solution.

I labelled all tests using `useDocument` option as `useDocument.test` and excluded them in `testPathIgnorePatterns` Jest option. Labelling tests might be useful in future refactoring (even though unfortunately Jest doesn't support any actual tagging functionality).

toomuchdesign:
It seems we're also hitting a further TS issues originated by Next.js which will be adressed in v11.1.3:
https://github.com/vercel/next.js/issues/29071

toomuchdesign:
[Here](https://github.com/vercel/next.js/commit/08a24780b01608e16b7733d31bda2aaf36fd1750#diff-63632e6e580e69692941c89bab0a0e0436ade3f9543ef610925f795928d9d5fdL271) is how the previous `Document.renderDocument` call was replaced.

PR #267, opened 2021-09-23
fix: disable useDocument option

toomuchdesign:

Temporary bug fix

#263

## What is the new behaviour?

Disable `useDocument` option to allow users to use `next-page-tester` (without _document rendering)

Yes. `useDocument` will throw an error when set to `true`.

## Other information:

This is a temporary workaround. Document rendering logic is still there but never reached at runtime. For this reason code coverage is expected to decrease.

PR #264 is the place where to start from to try and fix document rendering.


toomuchdesign:
Haven't received feedback, yet.
If I don't hear any contrary opinion I'm considering releasing this PR in the weekend to contain the issue. Then we'll be able to focus on #263 / #267 (re-enabling document rendering) in a separate context.

PR #285, opened 2021-11-06
feat: implement getInitialProps server side & client side redirects

Meemaw:

## What is the new behaviour?

...

### DIFF
--- a/src/__tests__/page-data-fetching/page-data-fetching.test.tsx
+++ b/src/__tests__/page-data-fetching/page-data-fetching.test.tsx
@@ -19,6 +19,10 @@ describe('Data fetching', () => {
       const expectedQuery = { foo: 'bar' };
 
       const { nextRoot: actual } = render();
+
+      const res = httpMocks.createResponse();
+      // @ts-expect-error missing type
+      res._eventsCount = 1;
       const expectedContext = {
         pathname: '/gip/[id]',
         query: { ...expectedParams, ...expectedQuery },
@@ -29,7 +33,7 @@ describe('Data fetching', () => {
           params: expectedParams,
           query: expectedQuery,
         }),
-        res: httpMocks.createResponse(),
+        res,
         err: undefined,
       };
 
@@ -51,6 +55,10 @@ describe('Data fetching', () => {
       const expectedQuery = { foo: 'bar' };
 
       const { nextRoot: actual } = render();
+
+      const res = httpMocks.createResponse();
+      // @ts-expect-error missing type
+      res._eventsCount = 1;
       const expectedProps = {
         params: expectedParams,
         query: expectedQuery,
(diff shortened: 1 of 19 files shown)


Meemaw:
Hi @toomuchdesign , long time no see 🙂 

I have a PR here for a feature I'm personally really excited about. Note that I'm leaving it in draft to get some early feedback, before I write tests for 100% test coverage and add documentation updates. The code however is already pretty much finalised IMO.

toomuchdesign:
Hi @Meemaw, glad to see you around :)
I took a quick ride through your PR and I was wondering:

- Is returning a redirect status code from within `getInitialProps` an intrinsic Next.js feature?
- Is it implemented both in app and page `getInitialProps`?

I'll try to go deeper as soon as I can. Thanks!

I have the feeling that overall complexity is crossing a threshold that's hard to keep under control. What are your feelings?

Also, it seems there are some major issues with Next 12 which I haven't been able to investigate, yet. Do you projects stick to v11?

Have a bright day!

Meemaw:
- Its not retuning statusCode from `getInitialProps` but its writing to http response directly. This is the only way to redirect on server when using `getInitialProps` and essentially only way to do redirects in `_app`.
- Yes, its implemented for both `_app` and `page`

While I initially though doing this will be super complicated, I'm actually quite happy with the outcome. The code changes are quite minimal in my opinion and we're reusing the code for doing redirects with `GSSP`.

---
I think this feature is super useful as doing authentication in `_app` is quite common and apps usually redirect to a `/login`  page from within. If you think this adds too much complexity to this project we can always fork the repo and use that, but I think this feature would be beneficial to other people.

---
We're mostly still on NextJS v11 due to various problems so we're sticking with that for now. I think most of the bigger nextjs applications have problems migrating to v12.

toomuchdesign:
Hi @Meemaw,
I started a review but unfortunately I can hardly follow the line. This is because I haven't touched Next.js for over a year and I'm definitely out of this context. As long as I see green tests and I trust you I'm willing to approve and merge, if the case. 👍 

This opens to a maintenance topic that I'd like to make public and discuss with you. The health of this project seems becoming more critical month by month, and -as I maintainer- I wouldn't be able to fix/update a feature like the one implemented in this PR which I recognize being very valuable.

Also I'd like not to be blocker to anybody willing to bring `next-page-tester` further.

I'd love discuss with you what options we have. Could moving `next-page-tester` to its own GitHub org improve the situation? Maybe I should open an issue to give this discussion more visibility.

Have a bright evening!

Meemaw:
@toomuchdesign I understand the pain points of maintaining this, and appreciate all the work 🙏 

1. I think opening an issue to open up some discussion around the potential org, or something similar would be the best. If people are really using this within their companies (ant not only for side projects), they will have incentive to push this forward. It shouldn't be only on you to maintain this. For example, a need for this feature arose in our project, so I've added the implementation in this PR, and since we will rely on it, I'll also have to own & maintain it.
2.Actually, I've been in talks with the Vercel folk recently and I've expressed my frustrations with the lack of such library from their side, which could really benefit the community. They heard me, so perhaps something will happen here.

PR #294, opened 2022-02-23
Next v12 compatibility

jasonwilliams:

Compatibility with next v12

There were 2 type issues, one was this:
https://github.com/vercel/next.js/pull/28646
Which needed defaultGetInitialProps to be added.

The other resolved itself by having [useMaybeDeferContent removed](https://github.com/vercel/next.js/pull/34506).

How to test this PR:
- You will need to be on node v16+
- You will need to be on Next v12+
- `npm i --save-dev next-page-tester/next-page-tester#pull/294/head`
- Set `__NEXT_TEST_MODE=jest` env

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,0 @@
-{
-  "presets": ["next/babel"],
-  "plugins": []
-}
--- a/package-lock.json
+++ b/package-lock.json
@@ -65,13 +65,14 @@
         "styletron-engine-atomic": "^1.4.8",
         "styletron-react": "^6.0.2",
         "ts-jest": "^27.0.3",
-        "typescript": "^4.4.4"
+        "typescript": "^4.4.4",
+        "vitest": "0.5.5"
       },
       "engines": {
         "node": ">=v12.22.8"
       },
       "peerDependencies": {
-        "next": "^10.0.8 || ^11",
+        "next": "^10.0.8 || ^11 || ^12",
         "react": "^16.6.0 || ^17",
         "react-dom": "^16.6.0 || ^17"
       }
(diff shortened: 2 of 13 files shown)


jasonwilliams:
So the plan is to try and make this work with vitest instead of Jest.
This does, however, mean changing to use import() instead of require here: https://github.com/next-page-tester/next-page-tester/blob/master/src/loadFile.ts#L18

Which means `await` changes all over the place. Any help would be useful

jasonwilliams:
Funnily enough I don’t get the segmentation fault errors when switching to dynamic import instead of require (for load file). So this is good news. But the tests do fail, so will need to see what’s going on there

jasonwilliams:
Help is needed to run these tests and try to fix the failing ones. I think using import can help prevent the require issues. If not we may have to just move to vitest completely.

toomuchdesign:
Hi @jasonwilliams! If the one running here is the last version, I can see a segmentation error [on Babel compilation](https://github.com/next-page-tester/next-page-tester/runs/5338028515?check_suite_focus=true#step:4:43).

jasonwilliams:
The hanging tests have (kinda) been fixed.
I noticed this:
https://github.com/vercel/next.js/blob/canary/packages/next/server/config.ts#L644-L648

If you set the env var `__NEXT_TEST_MODE ` to `jest` before running the tests they will now all pass.

As a "quick fix" we could just do this and call it a day, but we would end up stuck with jest. That being said we may not have any other choice. I can take a look at vitest but if vitest doesn't work we have a path forward

jasonwilliams:
Forcing __NEXT_TEST_MODE to be 'jest' would fix this PR for all those who use Jest as a test runner. But i don't know the effects for Mocha etc..

toomuchdesign:
Ehi @jasonwilliams, I'll answer here to the question you asked privately about using [Jest's `isolateModules` ](https://jestjs.io/docs/jest-object#jestisolatemodulesfn)to feed the discussion. That was introduced to simulate Next.js server and client renders in a single testing environment. It turned out to be quite necessary to any non-basic Next setup.

You can read more about it [here](https://github.com/next-page-tester/next-page-tester/blob/master/docs/a-note-about-client-and-server-renders.md).

